PR #3, opened 2016-05-28
Added BlockLocator.

zuyu:
As a part of Data Exchange service, we need the `BlockLocator` to determine the block locations, and retrieve the network address of peers.


query_execution/QueryExecutionTypedefs.hpp
```diff
@@ -66,6 +67,19 @@ enum QueryExecutionMessageType : message_type_id {
   kRebuildWorkOrderMessage,  // From Foreman to Worker.
   kRebuildWorkOrderCompleteMessage,  // From Worker to Foreman.
   kPoisonMessage,  // From the CLI shell to Foreman, then from Foreman to Workers.
+
```
pateljm: Would you consider sorting this list to improve readability? 

I know the previous enum was also not sorted and probably should be too.

zuyu: This list is intend to sort in the order of the life cycle of a query. So do the block locator messages, from registration, block adding / deleting / locating, to unregistration.

PR #5, opened 2016-05-30
DO NOT MERGE: CI w/ gRPC To Test the Distributed Version.

zuyu:


pateljm:
@zuyu Since we have have protobuf under third_party, should we just go and upgrade that and not have such a specific dependency?

zuyu:
This is just a hack until the GRPC and Protobuf 3.0 becomes GA. And at that time, I bet that a binary version would be available as an `apt` package.

On the other hand, if we upgrade the protobuf third_party to 3.0 beta, we may encounter some linking issues when `grpc` tries to links `protobuf`.

pateljm:
@zuyu: Ok so we now reduce the configurations that we try. This seems bad in general. I understand Travis is timing out. Do we know why?

zuyu:
There is the only one gcc test killed by the Travis.

https://travis-ci.org/apache/incubator-quickstep/jobs/133990216#L5810

If we could somehow run it with 2 concurrent jobs like the rest tests, I think it would be fine. Otherwise, we may have to wait for that both the binary releases become available.

zuyu:
@navsan Do you have any insights regarding the clang linking issues? Thanks!

navsan:
It looks like clang can't find the standard libraries. Could you try -lstdc++ (or -lc++ if libc++ is installed)?

zuyu:
@navsan We need to fix the `cmake` error regarding `gflags-shared` (#13) when `ENABLE_DISTRIBUTED=ON`:

https://travis-ci.org/apache/incubator-quickstep/jobs/137651757#L3727

navsan:
I’m not sure why the gflags-shared target already exists. Is it because gRPC has its own gflags? If so, how does the static target work? 

@Zuyu, I don’t think I’ll be able to figure this out on my own.

zuyu:
@navsan I think it is related to how `tmb` explicitly specifies the library `gflags_nothreads-static`. And disabling the shared build works using `gcc`, although there are two cases that were killed due to time out.

FYI, linking w/ `libstdc++` does not fix the linking issues regarding `std::string` when using `clang`.

PR #6, opened 2016-05-31
DO NOT MERGE: Minor Improvements in Storage (WIP).

zuyu:
This small PR replaced depreciated macros with `glog`, and also chose to check the valid proto only in the debug build.

### DIFF
--- a/storage/BasicColumnStoreTupleStorageSubBlock.cpp
+++ b/storage/BasicColumnStoreTupleStorageSubBlock.cpp
@@ -1,6 +1,6 @@
 /**
  *   Copyright 2011-2015 Quickstep Technologies LLC.
- *   Copyright 2015 Pivotal Software, Inc.
+ *   Copyright 2015-2016 Pivotal Software, Inc.
  *
  *   Licensed under the Apache License, Version 2.0 (the "License");
  *   you may not use this file except in compliance with the License.
@@ -51,6 +51,8 @@
 #include "utility/PtrMap.hpp"
 #include "utility/PtrVector.hpp"
 #include "utility/ScopedBuffer.hpp"
+
+#include "glog/logging.h"
 
 using std::memcpy;
 using std::memmove;
@@ -106,9 +108,9 @@ BasicColumnStoreTupleStorageSubBlock::BasicColumnStoreTupleStorageSubBlock(
                            sub_block_memory_size),
       sorted_(true),
       header_(static_cast<BasicColumnStoreHeader*>(sub_block_memory)) {
-  if (!DescriptionIsValid(relation_, description_)) {
-    FATAL_ERROR("Attempted to construct a BasicColumnStoreTupleStorageSubBlock from an invalid description.");
-  }
+  CHECK(DescriptionIsValid(relation_, description_))
+      << "Attempted to construct a BasicColumnStoreTupleStorageSubBlock from an invalid description:\n"
+      << description_.DebugString();
 
   sort_column_id_ = description_.GetExtension(BasicColumnStoreTupleStorageSubBlockDescription::sort_attribute_id);
   sort_column_type_ = &(relation_.getAttributeById(sort_column_id_)->getType());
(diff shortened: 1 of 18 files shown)


zuyu:
CI failed due to the failure to fetch llvm compiler.

storage/BasicColumnStoreTupleStorageSubBlock.cpp
```diff
@@ -582,15 +584,14 @@ TupleIdSequence* BasicColumnStoreTupleStorageSubBlock::getMatchesForPredicate(
...
 
-  if (matches == nullptr) {
-    FATAL_ERROR("Called BasicColumnStoreTupleStorageSubBlock::getMatchesForPredicate() "
-                "with a predicate that can only be evaluated with a scan.");
-  } else {
-    if (filter != nullptr) {
-      matches->intersectWith(*filter);
-    }
-    return matches;
+  DCHECK(matches != nullptr)
```
hbdeshmukh: This should be a ``CHECK`` instead of ``DCHECK``.

storage/BloomFilterIndexSubBlock.cpp
```diff
@@ -58,8 +58,9 @@ BloomFilterIndexSubBlock::BloomFilterIndexSubBlock(const TupleStorageSubBlock &t
       random_seed_(kBloomFilterSeed),
       bit_array_size_in_bytes_(description.GetExtension(
                                    BloomFilterIndexSubBlockDescription::bloom_filter_size)) {
-  CHECK(DescriptionIsValid(relation_, description_))
-      << "Attempted to construct an BloomFilterIndexSubBlock from an invalid description.";
+  DCHECK(DescriptionIsValid(relation_, description_))
```
hbdeshmukh: Why should we change a ``CHECK`` to a ``DCHECK``?

hbdeshmukh:
I am not sure if a conversation on a GitHub PR is acceptable. So I will not add more comments for now. There seem to be some inconsistencies while using ``DCHECK`` and ``CHECK`` macros in the code change. Please let me know once you address them.

zuyu:
Comments on GitHub are fine.

I use `DCHECK` generally for performance reasons, whenever the debug mode is good enough to ensure it will be ok as well in the release one. For example, `DescriptionIsValid` would be a debug check only.

However, there are some cases where we have to use `CHECK`, i.e., the expression within the `CHECK` needs to run in both modes (say `SerializeToArray`).

hbdeshmukh:
That's right. But in some cases, using ``DCHECK`` instead of ``CHECK`` may interfere with the correctness of the code. Please refer to the comment in storage/BasicColumnStoreTupleStorageSubBlock.cpp

storage/StorageBlock.cpp
```diff
@@ -91,57 +91,30 @@ StorageBlock::StorageBlock(const CatalogRelationSchema &relation,
...
   }
 
+  CHECK(block_header_.ParseFromArray(static_cast<char*>(block_memory_) + sizeof(int),
+                                     *static_cast<const int*>(block_memory_)))
+      << "A StorageBlockLayout created a malformed StorageBlockHeader.";
+
+  DCHECK(block_header_.IsInitialized());
+  DCHECK(StorageBlockLayout::DescriptionIsValid(relation_, block_header_.layout()));
+  DCHECK_EQ(block_header_.index_size_size(), block_header_.layout().index_description_size());
+  DCHECK_EQ(block_header_.index_size_size(), block_header_.index_consistent_size());
```
hbdeshmukh: Hi @zuyu 

I feel somewhat nervous regarding converting the above checks in this class from release mode to debug mode. Unless they are substantially expensive (which they are not), I don't think they should be made debug-only. Such checks can be very helpful if something goes wrong.

zuyu: My consideration for choosing `DCHECK` over `CHECK` would be, if a bug could reproduce in the debug mode, we don't need such check enabling for the release mode.

And `StorageBlockLayout::DescriptionIsValid` is somewhat expensive operations involving multiple virtual function calls and internal checks.

pateljm: @zuyu You make a good point. But, since this check happens rarely, would you consider accommodating @hbdeshmukh note and making these DCHECKs. He does make a good point, and such checks were really important in weeding out hard bugs in the code in our recent benchmarking effort. I think at a later stage in the project, when we have beat the code enough we could go back to only using such checks in Debug mode. Thanks!

zuyu: This check, however, happens quite frequently, whenever we create or load a block.

pateljm: @zuyu -- good point, but since our blocks are large this is not too bad IMHO. There are so many changes happening in the block layout, that I think for the time being good to have this check. Makes life so much easier for those of us who are running large end-to-end-tests. Thanks!

PR #7, opened 2016-06-03
Bug fixed in SimpleScalarSeparateChainingHashTable::isFull()

hbdeshmukh:
- Earlier check for isFull() was that whether there is at least one empty bucket available for insertion.
- Now isFull() can check if there is enough space for adding specified number of buckets.
- To reproduce the bug behind this fix: Run TPC-H Q2 on SF100 dataset after
  running the \analyze command. The optimizer allocates smaller than
  necessary space for one of the hash joins. The resize() is triggered
  but never gets executed because of the restrictive isFull() condition.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,11 +77,11 @@ addons:
   apt:
     sources:
       - ubuntu-toolchain-r-test
-      - llvm-toolchain-precise-3.7
+#      - llvm-toolchain-precise-3.7
     packages:
       - gcc-5
       - g++-5
-      - clang-3.7
+#      - clang-3.7
       - binutils-gold
       - libprotobuf-dev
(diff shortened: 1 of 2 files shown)


pateljm:
Looks good @hbdeshmukh. Merged. I know clang is failing, but that appears to be related to the ongoing LLVM issue.

zuyu:
@pateljm But the second commit is not related to this PR.

pateljm:
@zuyu You are right, and we can undo that if needed. You are also right that in the future, we must be careful to keep PRs focused.

zuyu:
@hbdeshmukh @pateljm I've dropped the second commit using the command line. So no worry about that, and use `git rebase -i` for merging new commits.

PR #8, opened 2016-06-06
Added query id in all the operators and the workorders

hbdeshmukh:
This PR adds the query ID field to all the relational operators and the workorder classes. This addition is going to be helpful in order to support running concurrent queries in Quickstep.


zuyu:
@hbdeshmukh What do you think if we put `query_id` as the first argument for every operator? Thanks!

relational_operators/WorkOrder.proto
```diff
@@ -45,6 +45,7 @@ enum WorkOrderType {
 
 message WorkOrder {
   required WorkOrderType work_order_type = 1;
+  required uint32 query_id = 2;
```
zuyu: We need to use `uint64`, as we use `std::size_t` to represent `query_id`.

relational_operators/WorkOrder.hpp
```diff
@@ -286,8 +286,15 @@ class WorkOrder {
   }
 
  protected:
-  WorkOrder() {}
+  /**
+   * @brief Constructor.
+   *
+   * @param query_id The ID of the query to which this WorkOrder belongs.
+   **/
+  explicit WorkOrder(const std::size_t query_id = 0)
```
zuyu: What do you think if we removed the default value for `query_id`.

hbdeshmukh:
Hi @zuyu 

I have incorporated your suggestions except the one about reordering ``query_id`` in the constructor arguments. I feel it is too much work for just a cosmetic refactoring. Please take a look again.

pateljm:
@hbdeshmukh: Zuyu (@zuyu) makes an excellent point, IMHO. The code would be more readable if we put the query_id first. This argument is somewhat inconsistently placed in the argument list  right now (e.g. in `DropTableOperator` it is the second argument, and in most other it is the last). 

@hbdeshmukh: How about doing this in a separate PR if it is too much to do it in this one? But, if possible, would be good to create that PR soon. I can pair with you if needed.

hbdeshmukh:
Hi @pateljm @zuyu I beg to differ on this point. In most places I tried to conform to the Google C++ style guide which we use as a reference. The relevant section in the style guide can be found here: https://google.github.io/styleguide/cppguide.html#Function_Parameter_Ordering 
It says that parameters should be ordered as inputs first (const references, const ints etc) followed by outputs (roughly speaking - pointers). Few places where I couldn't follow this convention was where already the convention wasn't followed. 

I think this issue boils down to personal style preferences that are not present in the style guide. Such comments are difficult to argue and enforce unlike some of the performance or correctness related comments which can be fairly objective. Going by such a rule on parameter ordering, I am sure there will be a lot of other places in the code which will need a refactoring. It will be better if either we modify the style guide so that everybody is on the same page or alternatively not enforce such a style preference.

zuyu:
@hbdeshmukh Ordering the input arguments is not only a code style thing, but putting `query_id` the first is somewhat the logical top-down flow from a query-level info down to a operator-level.

pateljm:
@hbdeshmukh Ok -- I see your point that the style guide contradicts putting the query_id first. How about we close this PR, and have a separate discussion on changes we need to make to the style guide. @zuyu Any comments on small change to the style guide to allow these changes?

zuyu:
@pateljm FYI, the style guide does not specify the order of any input arguments, including `query_id`.

pateljm:
@zuyu Good point. @hbdeshmukh and I discussed this and we agree with you. Will open a separate PR to improve the code quality. Thanks both @hbdeshmukh and @zuyu for your patience!

PR #9, opened 2016-06-08
DO NOT MERGE: Test an update to .travis.yml to get the clang coverage back.

pateljm:
Try using the new trusty version to get llvm working again. I understand that this could be slower, but allows us to keep the clang coverage tests when merging PRs.

Recall the ongoing issues with llvm.org as noted at http://lists.llvm.org/pipermail/llvm-dev/2016-May/100303.html


pateljm:
This PR is no longer needed given @navsan work in PR #12. Thanks Nav!

PR #10, opened 2016-06-08
Reorder query id parameter in various work order and operator constructors

hbdeshmukh:
This is a follow up of the review comments in PR #8. The query ID parameter used in various constructors of relational operators and work orders is the first parameter.

### DIFF
--- a/query_optimizer/ExecutionGenerator.cpp
+++ b/query_optimizer/ExecutionGenerator.cpp
@@ -193,8 +193,8 @@ void ExecutionGenerator::generatePlan(const P::PhysicalPtr &physical_plan) {
     }
     const QueryPlan::DAGNodeIndex drop_table_index =
         execution_plan_->addRelationalOperator(
-            new DropTableOperator(*temporary_relation,
-                                  query_handle_->query_id(),
+            new DropTableOperator(query_handle_->query_id(),
+                                  *temporary_relation,
                                   optimizer_context_->catalog_database(),
                                   false /* only_drop_blocks */));
     DCHECK(!temporary_relation_info.isStoredRelation());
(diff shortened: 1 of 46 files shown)


query_optimizer/ExecutionGenerator.cpp
```diff
@@ -1499,13 +1504,13 @@ void ExecutionGenerator::convertSort(const P::SortPtr &physical_sort) {
...
-                                        sort_run_gen_config_id,
-                                        input_relation_info->isStoredRelation(),
-                                        query_handle_->query_id()));
+      execution_plan_->addRelationalOperator(new SortRunGenerationOperator(
+          query_handle_->query_id(),
+          *input_relation_info->relation,
+          *initial_runs_relation,
+          initial_runs_destination_id,
+          sort_run_gen_config_id,
+          input_relation_info->isStoredRelation()));
```
zuyu: FYI, the original style of wrapping a line is acceptable based on our code style with at most 120 characters per line.

relational_operators/TextScanOperator.cpp
```diff
@@ -234,11 +234,10 @@ void TextScanOperator::receiveFeedbackMessage(const WorkOrder::FeedbackMessage &
   }
 }
 
-
```
zuyu: Please reset this change, as we use two lines to separate methods from different classes. In this case, a operator and its work order. Thanks.

query_optimizer/tests/ExecutionHeuristics_unittest.cpp
```diff
@@ -75,12 +75,12 @@ class ExecutionHeuristicsTest : public ::testing::Test {
     std::vector<attribute_id> build_attribute_ids;
     build_attribute_ids.push_back(build_attribute_id);
     QueryPlan::DAGNodeIndex build_operator_index =
-        query_plan->addRelationalOperator(new BuildHashOperator(*build_relation,
+        query_plan->addRelationalOperator(new BuildHashOperator(0,  /* dummy query ID */
```
zuyu: It is up to you to take the following suggestion.

In each unit test, we could have a const member called `query_id`, and pass it to each operator and its work orders.

And it is better to explicitly use a variable to represent the same query id value shared by all the operators, instead of multiple query id `0`s.

zuyu:
@hbdeshmukh I have two comments, including an optional. Thanks!

relational_operators/TextScanOperator.cpp
hbdeshmukh: I am not aware of this convention. It seems that many other relational operators don't follow it either. Did we document it somewhere?

zuyu: I cannot find it from our code style. But I remember I learnt it from some code reviews.

We can ignore this for now, and I would propose to add this to our code style. Thanks!

query_optimizer/tests/ExecutionHeuristics_unittest.cpp
zuyu: What do you think of this one? If you prefer as is, I could merge this PR.

hbdeshmukh: This suggestion will require some work as there are a number of unit tests which need this change.  The dummy query ID flag is informative as of now. We can come back to it later, if that's fine.

zuyu:
LGTM.

hbdeshmukh:
Thanks for the review @zuyu

PR #11, opened 2016-06-08
Fix typos in CMakeLists.txt.

navsan:
Both the CMakeLists files in this PR refer to libraries that don't exist. expressions/CMakeLists refers to a library that has probably been renamed. For the other file, I can't find any *_AggregationHandleUtil library. Not sure why it's been working so far!


zuyu:
LGTM

PR #13, opened 2016-06-09
Add option to build using shared libraries.

navsan:
Modify the build process to allow optional use of shared libraries instead of the static libraries we use today. The benefits are 
- much smaller build directory size (I’m seeing < 5GB for release mode, as opposed to the usual 40GB with static libs)
- slightly faster build time (went from 18-20min to 14-15min on the CloudLab machine with 40 jobs)
- reduced memory consumption (average memory consumption during build went from 12GB to 8.5GB)

To use this, just turn on the CMake flag BUILD_SHARED_LIBS. Currently, we don't support the use of both shared and static libraries in the build, because that leads to some really nasty bugs with the common third party libraries (GFlags, GLog, ProtoBuf) due to double-linking. 
All tests passed locally (both with the flag turned on and off). 

Also, see the attached shell script [build.shared.sh.txt](https://github.com/apache/incubator-quickstep/files/305928/build.shared.sh.txt) (renamed to .txt to attach to this PR). In a clean build directory, it runs the build commands, outputs them to a log file with timestamps (so that we can profile the build process if we want to), records the memory and cpu load statistics periodically, and computes the time taken and the average memory consumption during the build.

All of the changes in the commit are to do with linking to gflags library, since its name depends on whether it's shared or static.

### DIFF
--- a/cli/CMakeLists.txt
+++ b/cli/CMakeLists.txt
@@ -30,6 +30,12 @@ if(LIBNUMA_FOUND)
   set(QUICKSTEP_HAVE_LIBNUMA TRUE)
 endif()
 
+if (BUILD_SHARED_LIBS)
+  set(GFLAGS_LIB_NAME gflags_nothreads-shared)
+else()
+  set(GFLAGS_LIB_NAME gflags_nothreads-static)
+endif()
+
 configure_file (
   "${CMAKE_CURRENT_SOURCE_DIR}/CliConfig.h.in"
   "${CMAKE_CURRENT_BINARY_DIR}/CliConfig.h"
@@ -110,7 +116,7 @@ target_link_libraries(quickstep_cli_InputParserUtil
                       ${LIBNUMA_LIBRARY})
 endif()
 target_link_libraries(quickstep_cli_PrintToScreen
-                      gflags_nothreads-static
+                      ${GFLAGS_LIB_NAME}
                       quickstep_catalog_CatalogAttribute
                       quickstep_catalog_CatalogRelation
                       quickstep_storage_StorageBlock
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,13 +12,26 @@ cache: ccache
 
 compiler:
   - gcc
-  # clang
+  - clang
 
 env:
   - BUILD_TYPE=Debug VECTOR_COPY_ELISION_LEVEL=joinwithbinaryexpressions
   - BUILD_TYPE=Release VECTOR_COPY_ELISION_LEVEL=joinwithbinaryexpressions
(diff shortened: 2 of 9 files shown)


cli/CMakeLists.txt
```diff
@@ -30,6 +30,12 @@ if(LIBNUMA_FOUND)
   set(QUICKSTEP_HAVE_LIBNUMA TRUE)
 endif()
 
+if (BUILD_SHARED_LIBS)
+  set(GFLAGS_LIB_NAME gflags_nothreads-shared)
+else()
+  set(GFLAGS_LIB_NAME gflags_nothreads-static)
+endif()
```
zuyu: Try remove this part, as I believe `cmake` is smart enough to figure out the variable `GFLAGS_LIB_NAME`.

Otherwise, try use `options(GFLAGS_LIB_NAME ...)`.

navsan: You'd think so, but it's not! There's no common target gflags_nothreads that you can link against. It's only a problem with this particular third_party library. I think it's only fixed in the latest version of gflags (https://github.com/gflags/gflags/issues/117).

zuyu:
As for the double linking, try the following:

```
if (NOT TARGET gflags_nothreads-static)
  add_subdirectory("${THIRD_PARTY_SOURCE_DIR}/gflags" "${CMAKE_CURRENT_BINARY_DIR}/third_party/gflags")
  include_directories("${CMAKE_CURRENT_BINARY_DIR}/third_party/gflags/include")
endif()
```

navsan:
Yeah, I saw that we do exactly that in tmb. But we'd need to do this for each of the third_party libraries and that gets really messy. Figuring out which library is getting double-linked is extremely hard to do (I spent hours staring at the symbols in the libraries and executables to figure it out). So we'd best not support such a mixed compilation/linking option for now. It's not particularly useful anyway.

cli/CMakeLists.txt
zuyu: FYI, I've used both `gflags` and `glog` as git submodules for a private repo, with the condition `NOT TARGET gflags_nothreads-static` to choose whether compile `gflags`.

navsan: Yes, as I noted in the comment below, we can definitely do that. But 
- CMake is *not* smart enough to figure out the double-linking on its own.
- Specifically for gflags, there's no common target name that we can use to link whichever static/shared automatically so that CMake can resolve the library name on its own. This works for all other libraries, third_party and our own. Just not for gflags.
- The NOT TARGET test will have to be added everywhere that a third_party library is being linked, in all the CMakeLists. That's a *lot* of changes for very limited benefit.

PR #14, opened 2016-06-09
QUICKSTEP-8: Long running Foreman thread

hbdeshmukh:
This PR introduces basic skeleton of PolicyEnforcer and makes the Foreman thread understand concurrent queries, as well as allows the Foreman thread to live throughout the lifetime of the Quickstep process. Some key highlights are described below: 

- Foreman (C++ class) is now primarily responsible for dispatching work orders to workers and collect messages coming from worker threads during query execution. 

- Foreman thread now doesn't die after receipt of a new query, unlike earlier. Instead it stays alive throughout the lifetime of Quickstep process. 

- The main thread is now capable of sending and receiving messages through TMB. 

- A new query sent via command line is sent from the main thread to the Foreman thread via a TMB message. 

- Creation of a sekeleton PolicyEnforcer class.

- PolicyEnforcer can talk to multiple QueryManager instances to collect WorkOrders to be dispatched. 

- PolicyEnforcer receives messages from Foreman, figures out the recipient of the message (using query ID in the message) and relays it to the appropriate QueryManager instance. 

- PolicyEnforcer can also perform admission control. 

- Moving some commonly used code in utility class.


hbdeshmukh:
Apologies for the long list of commits in the PR. It is carried from the old Pivotal repo and a lot of commits  happened since then.

zuyu:
@hbdeshmukh Please do a `git rebase -i origin/master`. Otherwise, we will mess up the master again.

hbdeshmukh:
Hi @zuyu The rebase is complete.

cli/CommandExecutor.cpp
```diff
@@ -194,11 +198,14 @@ void executeDescribeTable(
...
  * @brief A helper function that executes a SQL query to obtain a scalar result.
  */
-inline TypedValue executeQueryForSingleResult(const std::string &query_string,
-                                               StorageManager *storage_manager,
-                                               QueryProcessor *query_processor,
-                                               SqlParserWrapper *parser_wrapper,
-                                               Foreman *foreman) {
+inline TypedValue executeQueryForSingleResult(
+    const tmb::client_id main_thread_client_id,
+    const tmb::client_id foreman_client_id,
```
zuyu: I would suggest to move `client_id` related arguments before `bus`, due to the same `tmb` context.

```
  const std::string &query_string,
  const tmb::client_id main_thread_client_id,
  const tmb::client_id foreman_client_id,
  tmb::MessageBus *bus,
```

cli/QuickstepCli.cpp
```diff
@@ -323,13 +326,20 @@ int main(int argc, char* argv[]) {
...
+                  query_processor->getDefaultDatabase(),
+                  query_processor->getStorageManager(),
+                  num_numa_nodes_system);
 
   // Start the worker threads.
   for (Worker &worker : workers) {
     worker.start();
   }
 
+  foreman.start();
```
zuyu: What do you think of moving this line before workers start, thus matching the reverse of the `join` order?

zuyu:
@hbdeshmukh Since we have a JIRA ticket for this, could we rename the title to `QUICKSTEP-8: Long running Foreman thread`?

zuyu:
@hbdeshmukh We have to rebase again, as we had "git merge" again.

storage/InsertDestination.cpp
```diff
@@ -60,25 +60,28 @@ InsertDestination::InsertDestination(const CatalogRelationSchema &relation,
                                      const StorageBlockLayout *layout,
                                      StorageManager *storage_manager,
                                      const std::size_t relational_op_index,
+                                     const std::size_t query_id,
```
zuyu: I suggest to move `query_id` before `relational_op_index` based on a top-down relations of things.

cli/QuickstepCli.cpp
hbdeshmukh: It shouldn't make a difference. The threads will be terminated anyway.

storage/InsertDestination.cpp
hbdeshmukh: I think this is really reading too much into ordering of parameters. It doesn't make any crucial difference.

cli/CommandExecutor.cpp
hbdeshmukh: Same as the InsertDestination comment, the ordering is really not that important here. If the doxygen is correct and the call sites follow the same order of arguments, we shouldn't be too bothered about it IMHO.

hbdeshmukh:
Hi @zuyu 

Replied to your comments. 

Please take a look at various places where query ID is now injected e.g query context, insert destination and their relevant proto formats. I have made it sure that the query ID passing happens correctly but If you have any concerns about that, I would be glad to know.

cli/QuickstepCli.cpp
zuyu: In the distributed version, `Foreman` starts first, and then `Worker`s starts to discover `Foreman`.

So, I believe this is the right flow for the single node as well, although in this case, `Foreman` knows `Worker`s in advance, and I admit that it does not affect the correctness. But to me it is the best practice to initialize and terminate threads.

@pateljm Any comments?

storage/InsertDestination.cpp
zuyu: The API design is crucial.

zuyu:
@hbdeshmukh I am curious about why we need `query_id` in `QueryContext`, although it does not hurt.

When we support multiple concurrent queries, we would have a data structure to keep trace of a mapping from `query_id` to `QueryContext`, right?

query_execution/QueryContext.hpp
```diff
@@ -472,6 +479,8 @@ class QueryContext {
   std::vector<std::unique_ptr<Tuple>> tuples_;
   std::vector<std::unordered_map<attribute_id, std::unique_ptr<const Scalar>>> update_groups_;
 
+  const std::size_t query_id_;
```
zuyu: This is my personal preference: put any `const` data members before any non-`const`.

But it is up to you.

zuyu:
FYI, In `optimizer/ExecutionGenerator`, we set `query_id` in `QueryContext` proto, not `QueryContext` itself.

So, I think in this PR, we don't need to add `query_id` in `QueryContext` object.

hbdeshmukh:
I am sorry, I take the part about InsertDestination back. I am indeed setting the query ID passed to the InsertDestination in ``ExecutionGenerator`` and not in ``QueryContext``. 

Sure, I can remove the query_id from the QueryContext class itself.

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -1635,5 +1635,11 @@ void ExecutionGenerator::convertTableGenerator(
   temporary_relation_info_vec_.emplace_back(tablegen_index, output_relation);
 }
 
+S::InsertDestination* ExecutionGenerator::addNewInsertDestinationToQueryContext() {
+  S::InsertDestination *insert_destination_proto(query_context_proto_->add_insert_destinations());
+  insert_destination_proto->set_query_id(query_context_proto_->query_id());
```
zuyu: I have an alternative implementation suggestions:

 1. Reset changes in this file and `InsertDestination.proto`.
 1. Modify `InsertDestination::ReconstructFromProto` to take a new argument `query_id` from `QueryContext proto`'s field `query_id`.

zuyu:
@hbdeshmukh I see the changes in `ExecutionGenerator` and `InsertDestination`, but to me, it is a overkill.

We could do it much simpler using the only `query_id` in `QueryContext` proto, as mentioned in a comment above. Thanks!

hbdeshmukh:
Hi @zuyu 

Reset the changes in ExecutionGenerator cpp and InsertDestination as per your suggestion. Please take another look.

query_optimizer/tests/ExecutionGeneratorTestRunner.cpp
```diff
@@ -90,13 +94,18 @@ void ExecutionGeneratorTestRunner::runTestCase(
...
+            main_thread_client_id_,
+            foreman_->getBusClientID(),
+            &query_handle,
+            &bus_);
+
+        // Receive workload completion message from Foreman.
+        const AnnotatedMessage annotated_msg =
+            bus_.Receive(main_thread_client_id_, 0, true);
+        const TaggedMessage &tagged_message = annotated_msg.tagged_message;
+        DCHECK(tagged_message.message_type() == kWorkloadCompletionMessage);
```
zuyu: Change to `DCHECK_EQ`?

query_execution/AdmitRequestMessage.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+  /**
+   * @brief Get the query handles from this message.
+   **/
+  const std::vector<QueryHandle*>& getQueryHandles() const {
+    return query_handles_;
+  }
+
+ private:
+  std::vector<QueryHandle*> query_handles_;
```
zuyu: Question:

How to use this pointer-based mechanism in the distributed `Quickstep`?

query_execution/CMakeLists.txt
```diff
@@ -235,36 +241,6 @@ if (ENABLE_DISTRIBUTED)
   add_test(BlockLocator_unittest BlockLocator_unittest)
 endif()
 
-add_executable(Foreman_unittest
-  "${CMAKE_CURRENT_SOURCE_DIR}/tests/Foreman_unittest.cpp")
```
zuyu: @hbdeshmukh Since we removed this test, please also remove its source file. Thanks!

query_execution/Foreman.cpp
```diff
@@ -22,355 +22,189 @@
...
 using std::move;
-using std::pair;
 using std::size_t;
+using std::unique_ptr;
 using std::vector;
 
 namespace quickstep {
 
-void Foreman::initialize() {
+DEFINE_uint64(min_load_per_worker, 2, "The minimum load defined as the number "
```
zuyu: @hbdeshmukh What do you think of moving this `gflag` variable in `ForemanLite`?

Thus, we could use it in the distributed version as well. Thanks!

query_execution/AdmitRequestMessage.hpp
zuyu: Never mind, both the optimizer and `Foreman` would be within the same process.

query_execution/Foreman.cpp
```diff
@@ -22,355 +22,189 @@
...
-      }
-    }
-    // Either no workorder found on the given NUMA node, or numa_node is -1.
-    // Try to get a normal WorkOrder from other NUMA nodes.
-    work_order = workorders_container_->getNormalWorkOrder(index);
-    if (work_order != nullptr) {
-      query_exec_state_->incrementNumQueuedWorkOrders(index);
-      return WorkerMessage::WorkOrderMessage(work_order, index);
+void Foreman::dispatchWorkerMessages(const vector<unique_ptr<WorkerMessage>> &messages) {
+  for (auto const &message : messages) {
```
zuyu: Please refactor to `const auto`.

query_execution/Foreman.cpp
```diff
@@ -22,355 +22,189 @@
...
+    DCHECK(message != nullptr);
+    int recipient_worker_thread_index = message->getRecipientHint();
```
zuyu: We could mark `const`.

query_execution/Foreman.cpp
```diff
@@ -22,355 +22,189 @@
...
+    if (recipient_worker_thread_index != -1) {
```
zuyu: Would we have some a `static constexpr` value, instead of a magic number `-1`?

query_execution/Foreman.hpp
```diff
@@ -22,46 +22,40 @@
...
-#include "glog/logging.h"
-#include "gtest/gtest_prod.h"
+#include "gflags/gflags.h"
 
+#include "tmb/id_typedefs.h"
 #include "tmb/message_bus.h"
 
 namespace quickstep {
 
+DECLARE_uint64(min_load_per_worker);
```
zuyu: Question: Does anywhere else also use this `gflag` variable?

query_execution/Foreman.hpp
```diff
@@ -71,233 +65,31 @@ class Foreman final : public ForemanLite {
...
-
  protected:
   /**
-   * @brief The foreman receives a DAG of relational operators, asks relational
-   *        operators to produce the workorders and based on the response it gets
-   *        pipelines the intermediate output to dependent relational operators.
-   *
-   * @note  The workers who get the messages from the Foreman execute and
-   *        subsequently delete the WorkOrder contained in the message.
+   * @brief Run the event-based loop in the Foreman thread.
```
zuyu: We could delete all the comments for a overridden method.

query_execution/PolicyEnforcer.cpp
```diff
@@ -0,0 +1,177 @@
...
+                           catalog_database_, storage_manager_, bus_));
+      return true;
+    } else {
+      LOG(ERROR) << "Query with the same ID " << query_id << " exists";
+      return false;
+    }
+  } else {
+    // This query will have to wait.
+    waiting_queries_.push(query_handle);
+    return false;
```
zuyu: This is some implementation tricks to save some indentation:

```
  if (...) {
    if (...) {
      ...
      return true;
    }

    LOG(ERROR) << ...;
    return false;
  }

  waiting_queries_.push(query_handle);
  return false;
```

query_execution/PolicyEnforcer.cpp
```diff
@@ -0,0 +1,177 @@
...
+  // TODO(harshad) - Make this function generic enough so that it
+  // works well when multiple queries are getting executed.
+  std::size_t per_query_share = 0;
+  if (!admitted_queries_.empty()) {
+    per_query_share = kMaxNumWorkerMessages / admitted_queries_.size();
+  } else {
+    LOG(WARNING) << "Requesting WorkerMessages when no query is running";
+    return;
+  }
+  DCHECK_GT(per_query_share, 0u);
```
zuyu: Some implementation tricks:

```
  if (admitted_queries_.empty()) {
    LOG(WARNING) << ...;
    return;
  }

  const std::size_t per_query_share = kMaxNumWorkerMessages / admitted_queries_.size();
  DCHECK_GT(per_query_share, 0u);
```

query_execution/PolicyEnforcer.cpp
```diff
@@ -0,0 +1,177 @@
...
+  std::vector<std::size_t> finished_queries_ids;
+
+  for (const auto &admitted_query_info : admitted_queries_) {
+    QueryManager *curr_query_manager = admitted_query_info.second.get();
+    DCHECK(curr_query_manager != nullptr);
+    std::size_t messages_collected_curr_query = 0;
+    while (messages_collected_curr_query < per_query_share) {
+      WorkerMessage *next_worker_message =
+          curr_query_manager->getNextWorkerMessage(0, -1);
```
zuyu: Please define the following constant to replace `-1`;

```
// catalog/CatalogTypedefs.hpp

constexpr numa_node_id kAnyNumaNodeId = -1;
```

query_execution/PolicyEnforcer.cpp
```diff
@@ -0,0 +1,177 @@
...
+        // Check if the query's execution is over.
+        if (curr_query_manager->getQueryExecutionState().hasQueryExecutionFinished()) {
+          // If the query has been executed, remove it.
+          finished_queries_ids.push_back(admitted_query_info.first);
+        }
+        break;
+      }
+    }
+  }
+  for (std::size_t finished_qid : finished_queries_ids) {
```
zuyu: Please add `const` before `std::size_t`. Thanks!

query_execution/PolicyEnforcer.hpp
```diff
@@ -0,0 +1,168 @@
...
+   * @brief Constructor.
+   *
+   * @param foreman_client_id The TMB client ID of the Foreman.
+   * @param num_numa_nodes Number of NUMA nodes used by the system.
+   * @param catalog_database The CatalogDatabase used.
+   * @param storage_manager The StorageManager used.
+   * @param bus The TMB.
+   **/
+  PolicyEnforcer(const tmb::client_id foreman_client_id,
+                 const std::size_t num_numa_nodes,
```
zuyu: Would we treat this argument as a optional one, with some default value? Thanks!

zuyu:
@hbdeshmukh Have you try to use `std::vector<const QueryHandle*>` instead of `std::vector<QueryHandle*>`?

I believe we don't need to change anything in `QueryHandle` after query optimization. Thanks!

query_execution/PolicyEnforcer.hpp
```diff
@@ -0,0 +1,168 @@
...
+   * @return True if there is at least one active or waiting query, false if
+   *         the policy enforcer doesn't have any query.
+   **/
+  inline bool hasQueries() const {
+    return !(admitted_queries_.empty() && waiting_queries_.empty());
+  }
+
+ private:
+  static constexpr std::size_t kMaxConcurrentQueries = 1;
+  static constexpr std::size_t kMaxNumWorkerMessages = 20;
```
zuyu: Instead of static variables, I would suggest to implement as `gflags` arguments.

query_execution/QueryContext.proto
```diff
@@ -54,4 +54,6 @@ message QueryContext {
 
   // NOTE(zuyu): For UpdateWorkOrder only.
   repeated UpdateGroup update_groups = 10;
+
+  required uint64 query_id = 11;
```
zuyu: Since we add a new field in `QueryContext` proto, we need to modify `QueryContext::ProtoIsValid` to reflect the change:

```
  return proto.IsInitialized();
```

query_execution/QueryExecutionMessages.proto
```diff
@@ -35,16 +36,19 @@ message CatalogRelationNewBlockMessage {
...
 message DataPipelineMessage {
   required uint64 operator_index = 1;
   required fixed64 block_id = 2;
   required int32 relation_id = 3;
+  required uint64 query_id = 4;
 }
 
 message WorkOrdersAvailableMessage {
   required uint64 operator_index = 1;
+  required uint64 query_id = 2;
```
zuyu: I would suggest to change to `query_id = 1` in each message based on the top-down flow from the query-level info to operator-level.

query_execution/QueryExecutionTypedefs.hpp
```diff
@@ -58,6 +58,7 @@ using ClientIDMap = ThreadIDBasedMap<client_id,
 
 // We sort the following message types in the order of a life cycle of a query.
 enum QueryExecutionMessageType : message_type_id {
+  kAdmitRequestMessage,  // Requesting a query (or queries) to be admitted.
```
zuyu: The comment would also state the message sender and receiver.

query_execution/QueryExecutionTypedefs.hpp
```diff
@@ -67,7 +68,8 @@ enum QueryExecutionMessageType : message_type_id {
                               // their corresponding RelationalOperators.
   kRebuildWorkOrderMessage,  // From Foreman to Worker.
   kRebuildWorkOrderCompleteMessage,  // From Worker to Foreman.
-  kPoisonMessage,  // From the CLI shell to Foreman, then from Foreman to Workers.
+  kWorkloadCompletionMessage,  // From Foreman to main thread.
+  kPoisonMessage,  // From the main thread to Foreman and workers.
```
zuyu: Please capitalize the first letter of `Worker`. Thanks!

query_execution/WorkOrdersContainer.hpp
```diff
@@ -76,7 +78,7 @@ class WorkOrdersContainer {
    * @return If there are pending WorkOrders.
    **/
   inline bool hasNormalWorkOrder(const std::size_t operator_index) const {
-    DEBUG_ASSERT(operator_index < num_operators_);
+    DCHECK(operator_index < num_operators_);
```
zuyu: Please use `DCHECK_LT`.

query_execution/WorkOrdersContainer.hpp
```diff
@@ -92,9 +94,9 @@ class WorkOrdersContainer {
    **/
   inline bool hasNormalWorkOrderForNUMANode(
       const std::size_t operator_index, const int numa_node_id) const {
-    DEBUG_ASSERT(operator_index < num_operators_);
-    DEBUG_ASSERT(numa_node_id >= 0);
-    DEBUG_ASSERT(static_cast<std::size_t>(numa_node_id) < num_numa_nodes_);
+    DCHECK(operator_index < num_operators_);
+    DCHECK_GE(numa_node_id, 0);
+    DCHECK(static_cast<std::size_t>(numa_node_id) < num_numa_nodes_);
```
zuyu: Please use `DCHECK_LT` instead of `DCHECK`.

query_execution/WorkerMessage.hpp
```diff
@@ -120,12 +137,13 @@ class WorkerMessage {
                 const WorkerMessageType type)
       : work_unit_(work_unit),
         relational_op_index_(relational_op_index),
-        type_(type) {
-  }
+        type_(type),
+        recipient_index_hint_(-1) {}
```
zuyu: Would we define sth like `kInvalidRecipientIndexHint = -1` and use it?

query_optimizer/ExecutionGenerator.hpp
```diff
@@ -374,6 +375,14 @@ class ExecutionGenerator {
    */
   void dropAllTemporaryRelations();
 
+  /**
+   * @brief Add a new InsertDesetination to the QueryContext and set the
+   *        query ID for the InsertDestination.
+   *
+   * @return A pointer to the serialized InsertDestination.
+   **/
+  serialization::InsertDestination* addNewInsertDestinationToQueryContext();
```
zuyu: Please remove this unused method. Thanks!

relational_operators/AggregationOperator.hpp
```diff
@@ -123,6 +123,8 @@ class AggregationWorkOrder : public WorkOrder {
    *
    * @param query_id The ID of this query.
    * @param input_block_id The block id.
+   * @param query_id The ID of this query.
+   * @param input_block_id The block id.
```
zuyu: Please remove these duplicated doxygen comments. I think they resulted from some incorrect auto-rebase.

relational_operators/DestroyHashOperator.hpp
```diff
@@ -83,6 +85,8 @@ class DestroyHashWorkOrder : public WorkOrder {
    *
    * @param query_id The ID of the query to which this WorkOrder belongs.
    * @param hash_table_index The index of the JoinHashTable in QueryContext.
+   * @param query_id The ID of the query to which this WorkOrder belongs.
+   * @param hash_table_index The index of the JoinHashTable in QueryContext.
```
zuyu: Ditto for the duplicated doxygen.

zuyu:
@hbdeshmukh I did one pass, and I would highlight two things:
 * Please remove some duplicated doxygen in operator header files, as a part of incorrect auto-rebasing.
 * Emphasis the argument order for `query_id` and `op_index`. We would always put the former first, and then latter.

Thanks!

cli/QuickstepCli.cpp
pateljm: @zuyu Your suggestion makes sense to make the distributed version easier. Thanks both of you for a good discussion!

pateljm:
@zuyu Thanks for a very thorough pass. Thanks @hbdeshmukh for your patience and working with these suggestions in advance!

query_execution/Foreman.cpp
hbdeshmukh: Are you suggesting that the same variable can be used by Foreman and ShiftBoss? If so, then it's better to have them defined in their individual files. Because clearly the thresholds will be different with respect to Foreman and ShiftBoss.

zuyu: No, only `ForemanDistributed`.

query_execution/PolicyEnforcer.hpp
hbdeshmukh: I have converted the kMaxNumWorkerMessages to a gflag variable. It will make more sense to make ``kMaxConcurrentQueries`` a gflag when we have a good support for running concurrent queries (i.e. various policies for selecting queries). Thanks for the suggestion.

query_execution/Foreman.cpp
hbdeshmukh: I don't understand this. How can it help if I move the gflag variable to ForemanLite? Both ForemanDistributed and ShiftBoss are the derived classes of the base class ForemanLite, correct? I am saying that the threshold described by this flag should be different for ForemanDistributed and ShiftBoss. In that case it makes more sense for both the classes to have their own gflag variable.

query_execution/Foreman.hpp
hbdeshmukh: In ``Foreman::canCollectNewMessages()``

query_execution/PolicyEnforcer.hpp
hbdeshmukh: Is there a reason why we should make it optional?

query_execution/QueryContext.proto
hbdeshmukh: Changed it, thanks for pointing it out.

storage/InsertDestination.cpp
pateljm: Ok -- how about in the sprit of moving forward, we close this discussion and PR if the reordering is the only issue. We can work on the reordering in a separate PR. 

@zuyu -- can you confirm that the reodering is the only issue that is left on the table from your perspective? Thanks.

query_execution/Foreman.cpp
zuyu: If you are curious about the distributed query execution, check out this branch: https://github.com/apache/incubator-quickstep/tree/distributed-exe-gen-test/query_execution.

In short, after moving the `gflag` variable to `ForemanLite`, `ForemanDistributed`, a derived class of `ForemanLite`, could use this variable. We use the term `Worker`, which consists of `Shiftboss`, `StorageManager`, and a thread pool, running as an independent `OS` process that executes `WorkOrder`s from `ForemanDistributed`. `Shiftboss` is a final derived class of `Thread`.

The load of `Worker` is controlled by `ForemanDistributed`, so it is perfect to set in `ForemanLite`.

query_execution/Foreman.hpp
zuyu: Then, we could remove this line. Since it is only used within the same class, we don't need to declare it in the header file.

Defining it in the source file is good enough.

hbdeshmukh: Removed.

hbdeshmukh:
> @hbdeshmukh Have you try to use std::vector<const QueryHandle*> instead of std::vector<QueryHandle*>?

Using ``const QueryHandle*`` doesn't work because we extract the DAG from it. On the DAG pointer, we call ``getNodePayloadMutable()`` to fetch a pointer to the ``RelationalOperator``. The function ``getAllWorkOrders`` changes the state of the DAG. Due to this chain, we can't use ``const QueryHandle*``.

pateljm:
LGTM. Merging. Let's handle any remaining issues re. reordering parameters in a separate PR.

query_execution/PolicyEnforcer.hpp
zuyu: Running `Quickstep` in a NUMA environment is optional, and the initial version of distributed prototype would not support `NUMA`.

PR #17, opened 2016-06-09
QUICKSTEP-10: WorkOrder Serialization.

zuyu:
This large PR serializes `WorkOrders` as a proto, as a part of achieving distributed version.

It also has some refactors for `TextScanOperator`s (#19).

Note that this would affect the on-going work for adding new operators like window aggregation.

### DIFF
--- a/query_execution/CMakeLists.txt
+++ b/query_execution/CMakeLists.txt
@@ -29,9 +29,10 @@ add_library(quickstep_queryexecution_QueryContext_proto
 add_library(quickstep_queryexecution_QueryExecutionMessages_proto
             ${queryexecution_QueryExecutionMessages_proto_srcs}
             ${queryexecution_QueryExecutionMessages_proto_hdrs})
-add_library(quickstep_queryexecution_QueryExecutionState ../empty_src.cpp QueryExecutionState.hpp) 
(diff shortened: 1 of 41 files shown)


pateljm:
@zuyu -- Nice! LGTM. As you say, will have to rework the window aggregation branch later, but sounds like the changes are straightforward. To that effect, could you add a guide for what new operator writers need to follow w.r.t. the proto. This could go in ` query_execution/WorkOrderProtosContainer.hpp` (but you may have a better location for that comment). Ok -- to do that as a separate PR if you find that more convenient.   

Can you merge the conflict -- sorry that may have been caused by the merge of PR#19.

pateljm:
Good suggestion @navsan ! Trying it now.

zuyu:
@pateljm This PR is ready to merge, and has sync-ed with the master. Please take another look. Thanks!
As for adding a document regarding how to add a new operator, we should do so in another separate PR, and write it in a `markdown` file in the `relational_operator` directory.

@jianqiao I did some refactors for `TextScanOperator`:
 * Rewrote the code that generates `TextScanWorkOrder`s.
 * Use a `gflags` parameter for the const value `textscan_text_segment_size`.

zuyu:
Interestingly, it was closed, but merged into master. But I did not do anything...

PR #19, opened 2016-06-09
Improve text scan operator

jianqiao:
This PR updates the `TextScanOperator` to improve its performance.

There are three main changes:
(1) Pass `text_offset` and `text_segment_size` as parameters to each `TextScanWorkOrder` instead of really loading the data. Then each `TextScanWorkOrder` reads the corresponding piece of data directly from disk.
(2) Avoid extra string copying by passing `const char **` buffer pointers into `parseRow()` and `extractFieldString()`.
(3) Use `ColumnVectorsValueAccessor` as the temporary container to store the parsed tuples. Then call `output_destination_->bulkInsertTuples()` to bulk insert the tuples.

**Note 1:** This updated version follows the semantics of the old `TextScanOperator` except that it does not support the backslash + newline escaping, e.g.
(a)
```
aaaa\
bbbb
```
which is semantically equivalent to
(b)
```
aaaa\nbbbb
```
The updated version supports (b) but not (a). As (a) incurs extra logic that complicates code. Meanwhile, format (a) seems to be specific to PostgreSQL, and the [documentation](http://www.postgresql.org/docs/9.6/static/sql-copy.html) of PostgreSQL 9.6 says:
_It is strongly recommended that applications generating COPY data convert data newlines and carriage returns to the \n and \r sequences respectively. At present it is possible to represent a data carriage return by a backslash and carriage return, and to represent a data newline by a backslash and newline. However, these representations might not be accepted in future releases. They are also highly vulnerable to corruption if the COPY file is transferred across different machines (for example, from Unix to Windows or vice versa)._

**Note 2:** This PR relies on the fix from #18 to work correctly for loading `TYPE compressed_columnstore` tables.

**Note 3:** Using 40 workers, the expected loading time on cloudlab machines with current SQL-benchmark settings are ~465s for SSB SF100 and ~1050s for TPCH SF100 (`*.tbl` files on slowdisk).


pateljm:
@jianqiao This is fantastic! This is going to making a huge difference when loading large datasets. LGTM. As a perspective loading SSB used to take about 2800 seconds. Huge 6X improvement!

Merging.

relational_operators/TextScanOperator.cpp
```diff
@@ -155,116 +63,50 @@ bool TextScanOperator::getAllWorkOrders(
...
-               text_blob_queue_.empty();
-      }
-    }
-    return false;
-  } else {
-    // Serial implementation.
-    if (blocking_dependencies_met_ && !work_generated_) {
-      for (const auto &file : files) {
+  // Text segment size set to 256KB.
+  constexpr std::size_t kTextSegmentSize = 0x40000u;
```
zuyu: @jianqiao I was wondering if there is a reason to choose this size. Thanks!

relational_operators/TextScanOperator.cpp
```diff
@@ -155,116 +63,50 @@ bool TextScanOperator::getAllWorkOrders(
...
             new TextScanWorkOrder(query_id_,
                                   file,
+                                  text_offset,
+                                  std::min(kTextSegmentSize, file_size - text_offset),
                                   field_terminator_,
                                   process_escape_sequences_,
                                   output_destination,
                                   storage_manager),
             op_index_);
+        text_offset += kTextSegmentSize;
```
zuyu: This won't become a bug, but I think what we really mean is the following:

```
  const size_t text_actual_segment_size = std::min(kTextSegmentSize, file_size - text_offset);
  text_offset += text_actual_segment_size;
```

relational_operators/TextScanOperator.cpp
```diff
@@ -274,439 +116,293 @@ TextScanWorkOrder::TextScanWorkOrder(const std::size_t query_id,
...
   const CatalogRelationSchema &relation = output_destination_->getRelation();
+  std::vector<Tuple> tuples;
 
-  string current_row_string;
-  if (is_file_) {
-    FILE *file = std::fopen(filename_.c_str(), "r");
-    if (file == nullptr) {
-      throw TextScanReadError(filename_);
-    }
+  constexpr std::size_t kSmallBufferSize = 0x4000;
```
zuyu: Again, I was curious how we came up with this number. Does a different value would change the loading performance?

relational_operators/TextScanOperator.cpp
```diff
@@ -274,439 +116,293 @@ TextScanWorkOrder::TextScanWorkOrder(const std::size_t query_id,
...
-      }
-    } while (have_row);
-
-    // Drop the consumed blob produced by TextSplitWorkOrder.
-    blob.release();
-    storage_manager_->deleteBlockOrBlobFile(text_blob_);
+  // Read text segment into buffer.
+  FILE *file = std::fopen(filename_.c_str(), "rb");
+  std::fseek(file, text_offset_, SEEK_SET);
+  std::size_t bytes_read = std::fread(buffer, 1, text_segment_size_, file);
```
zuyu: Please mark `const`.

jianqiao:
From experiments this is the best size for both _slowdisk_ and _fastdisk_ on the cloudlab machines. The observation is that _slowdisk_ loading favors smaller (256K) segment size and _fastdisk_ loading favors larger (1M~4M, but 256K is okay) segment size. I haven't dived into the underlying details though.

relational_operators/TextScanOperator.cpp
jianqiao: From experiments this is the best size for both _slowdisk_ and _fastdisk_ on the cloudlab machines. The observation is that _slowdisk_ loading favors smaller (256K) segment size and _fastdisk_ loading favors larger (1M~4M, but 256K is okay) segment size. I haven't dived into the underlying details though.

relational_operators/TextScanOperator.cpp
jianqiao: From experiments this is the best size for both _slowdisk_ and _fastdisk_ on the cloudlab machines. The observation is that _slowdisk_ loading favors smaller (256K) segment size and _fastdisk_ loading favors larger (1M~4M, but 256K is okay) segment size. I haven't dived into the underlying details though.

zuyu: Cool. I would suggest to use a gflags to define this value, so that we could set any value w/o recompilation.

pateljm:
@zuyu -- I think it is perfectly fine for you go clean it up and open a new PR.

relational_operators/TextScanOperator.cpp
jianqiao: This is the buffer size for processing the last row of the text segment.

For each text segment, we will first: (1) start scanning from the first newline (`\n`) character in the segment, and end scanning with the last newline character in the segment; and then: (2) scanning from the _last_ newline character in _this_ text segment to the _first_ newline character in the _next_ text segment (corner cases will also be handled).

Consider (2), how much data from the _next_ segment do we want to load from disk? Since it is just one row, in most cases we may not want to load too much. So the load buffer starts with 1024 bytes, and we keep appending the buffer's contents to a `std::string` if `\n` is not met. If this "tail row" is really large, the buffer will grow up to 0x4000 bytes.

relational_operators/TextScanOperator.cpp
jianqiao: Yes alternatively we can do
```
std::size_t text_segment_size = std::min(kTextSegmentSize, file_size - text_offset);

... // addNormalWorkOrder

text_offset += text_segment_size;
```

relational_operators/TextScanOperator.cpp
jianqiao: Yes it would be better to have a gflag on this.

PR #20, opened 2016-06-09
Change the vector elision from `none` to `selection`

pateljm:
The default VECTOR_COPY_ELISION_LEVEL is selection, so test that instead of none.


zuyu:
@pateljm How about trying with all the four options to see the time to take, and then go with there?

navsan:
GCC with VECTOR_COPY_ELISION_LEVEL=Selection failed in both debug and release modes. It's the same internal compiler error. 

Does the number of concurrently running jobs affect the amount of memory available to each? If so, we should definitely not bump up the number of build types.

zuyu:
No, the total memory would be the same. `gcc` maybe just do not spend memory resource wisely.

navsan:
@pateljm I think you should set MAKE_JOBS to 1 for Selection as well.

pateljm:
Good suggestion @navsan! Trying it now. Sad that `gcc` is fragile can can't deal with a parallel make.

zuyu:
@pateljm Before merging this PR, we need to rebase w/ master to get rid of previous merged commits.

zuyu:
@pateljm You might want to try add back the else case of `MAKE_JOBS=2` for `clang`.

pateljm:
Creating a new PR tied to JIRA.

PR #21, opened 2016-06-09
Use shared libraries when building on Travis

navsan:
This should reduce memory consumption and build time.


navsan:
Comparing Travis run times for this PR (https://travis-ci.org/apache/incubator-quickstep/builds/136474629) with one from earlier today (https://travis-ci.org/apache/incubator-quickstep/builds/136370961), almost all the run times have gone down to a certain extent. It's particularly helpful for the GCC runs with copy elision (>10% speedup).

PR #22, opened 2016-06-09
Add a shell script that logs the build process for later profiling

navsan:
Adding a utility script that I find helpful.

### DIFF
--- a/build/profile_build.sh
+++ b/build/profile_build.sh
@@ -1,0 +1,104 @@
+#!/bin/bash
+# This script may be useful for developers to profile the build process itself.
+#
+# This shell script runs CMake and make, dumping all output into a log file.
+# It also logs the CPU usage and memory information during the build.
+# All log messages are timestamped to enable profiling.
+#
+# Dependencies:
+# - ts
+# sudo apt-get install -y libtime-duration-perl moreutils
+# - vmstat and mpstat
+# sudo apt-get install -y sysstat
+#
+# Usage: ./profile_build.sh
+# Set the CMake and make command you want to use below.
+# If CMakeLists.txt is detected, the script exits because in-source-tree build
+# is not supported.
+# If CMakeCache.txt is detected, the script skips cmake and runs make only.
+
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+set -e
+
+CMAKE_COMMAND="cmake \
+-D BUILD_SHARED_LIBS=On \
+-D USE_TCMALLOC=0 \
+-D CMAKE_BUILD_TYPE=Debug \
+.. "
+MAKE_COMMAND="make VERBOSE=1"
+LOG_FILENAME=$LOG_FILENAME
+
+# Continuously dump memory usage and cpu load info to files for later analysis
+function start_stat_collectors {
+  rm -f stats_*.txt
+  vmstat -SM 3 | ts "%.s (%H:%M:%S)" > stats_mem.txt 3>&1 &
+  PID_vmstat=$!
+  mpstat 3 | ts "%.s (%H:%M:%S)" > stats_cpu.txt 2>&1  &
+  PID_mpstat=$!
+}
+
+function kill_stat_collectors {
+  kill $PID_vmstat
+  kill $PID_mpstat
+  exit
+}
+
+function check_directory {
+  if [[ -f CMakeLists.txt ]]; then
+    echo "Running the build in the source tree is not supported."
+    exit 1
+  fi
+}
+
+function log_repo_version_info {
+  git log master... >> $LOG_FILENAME
+  git diff master >> $LOG_FILENAME
+}
+
+function run_cmake {
+  if [[ ! -f CMakeCache.txt ]]; then
+    echo "$CMAKE_COMMAND" | tee -a $LOG_FILENAME
+    $CMAKE_COMMAND 2>&1 | ts "%.s (%H:%M:%S)" | tee -a $LOG_FILENAME
+  else
+    echo "CMakeCache.txt detected. Not running CMake again."
+  fi
+}
+
+function run_make {
+  echo "$MAKE_COMMAND" | tee -a $LOG_FILENAME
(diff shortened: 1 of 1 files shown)


pateljm:
LGTM. Merging.

build/profile_build.sh
```diff
@@ -0,0 +1,66 @@
...
+avg_mem=`grep -v r stats_mem.txt | tr -s ' ' | awk -F " " '{s+= $6; c++} END {print s/c/1024}'`
+echo
+echo
+echo "Average memory used was $avg_mem GB"  | tee -a build.log
+
+time_taken=`expr $END_TIME - $START_TIME`
+mins=`expr $time_taken / 60`
+secs=`expr $time_taken % 60`
+echo "Time taken was ${mins}m ${secs}s" | tee -a build.log
+
```
zuyu: Please remove the last empty line.

build/profile_build.sh
```diff
@@ -0,0 +1,66 @@
...
+# All log messages are timestamped to enable profiling.
+#
+# Dependencies:
+# - ts 
+# sudo apt-get install -y libtime-duration-perl moreutils
+# - vmstat and mpstat 
+# sudo apt-get install -y sysstat
+#
+# Usage:
+# Set the CMake flags you want to use below.
```
zuyu: Could we have the exact shell command to use this script? Thanks!

navsan:
Applied the suggested changes.

build/profile_build.sh
```diff
@@ -0,0 +1,65 @@
...
+vmstat -SM 3 | ts "%.s (%H:%M:%S)" > stats_mem.txt 3>&1 &
+PID_vmstat=$!
+mpstat 3 | ts "%.s (%H:%M:%S)" > stats_cpu.txt 2>&1  &
+PID_mpstat=$!
+
+START_TIME=`date +"%s"`
+if [[ ! -f CMakeCache.txt ]]; then 
+  CMAKE_COMMAND="cmake -D BUILD_SHARED_LIBS=On -D USE_TCMALLOC=0 \
+               -D CMAKE_BUILD_TYPE=Debug \
+               -D CMAKE_C_COMPILER=/usr/bin/clang -D CMAKE_CXX_COMPILER=/usr/bin/clang++ .. "
```
zuyu: I was wondering why we need to specify the compiler. What if we use `gcc` here?

navsan: Of course. That’s really only an example. See the Usage notes: you have to specify the CMake command you want to use.

build/profile_build.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+# This script may be useful for developers to profile the build process itself.
```
zuyu: We may need to add copyright for this file.

build/profile_build.sh
```diff
@@ -0,0 +1,65 @@
...
+# - vmstat and mpstat 
+# sudo apt-get install -y sysstat
+#
+# Usage: ./profile_build.sh [num_parallel_jobs]
+# The number of parallel make jobs can be passed as argument to this script.
+# By default, it is the number of (virtual) CPUs on the machine (nproc).
+# Set the CMake flags you want to use below.
+#
+# If CMakeCache.txt is detected, the script skips cmake and runs make only.
+#
```
zuyu: Please remove this comment line.

zuyu:
I just rebase it with the master, and I'll merge it after running it on a `Linux` box. Thanks!

zuyu:
@navsan This file has a lot of trailing whitespaces that need to remove.

build/profile_build.sh
```diff
@@ -0,0 +1,65 @@
...
+
+NUM_JOBS=`nproc`
+if [[ "$1" != "" ]]; then 
+    NUM_JOBS=$1
+fi
+
+echo "Starting build in " `pwd`  >> build.log
+echo
```
zuyu: ?

build/profile_build.sh
zuyu: So I suggest to remove this compiler settings.

In a default case, we may not assume that users have `clang`, particularly nowadays when `llvm.org/apt` is switched off.

build/profile_build.sh
navsan: Added a License.

zuyu:
Let me know once you want to merge it. Thanks!

navsan:
Sure. Just squashing commits, testing it out, etc. Thanks Zuyu!

build/profile_build.sh
```diff
@@ -0,0 +1,111 @@
...
+#   limitations under the License.
+
+set -e
+
+CMAKE_COMMAND="cmake \
+-D BUILD_SHARED_LIBS=On \
+-D USE_TCMALLOC=0 \
+-D CMAKE_BUILD_TYPE=Debug \
+.. "
+MAKE_COMMAND="make VERBOSE=1 -j `nproc`"
```
zuyu: We may need to give users the ability to overwrite the number of concurrent jobs to run `make`.

To me, in some `VM` environments the build process would be killed due to out of memory unless we use 1 job to build, although it supports `4` build jobs.

build/profile_build.sh
```diff
@@ -0,0 +1,111 @@
...
+# Usage: ./profile_build.sh
+# Set the CMake and make command you want to use below.
+# By default, the number of parallel make jobs is equal to the number of
+# (virtual) CPUs on the machine (nproc).
+# If CMakeLists.txt is detected, the script exits because in-source-tree build
+# is not supported.
+# If CMakeCache.txt is detected, the script skips cmake and runs make only.
+
+#   Copyright 2011-2015 Quickstep Technologies LLC.
+#   Copyright 2015-2016 Pivotal Software, Inc.
```
zuyu: Also, please remove above two copyright lines, as you are not a part of both.

build/profile_build.sh
navsan: Then just change it in the script. This is not meant to be a stand-alone tool with lots of configuration options. Just a helper/utility script to not have to re-invent the wheel. Just like NUM_JOBS, the build type, compiler, build target, etc. are all things one would want to customize.

build/profile_build.sh
navsan: Removed.

build/profile_build.sh
zuyu: I agree w/ ur point, and thus after removing `-j nproc` and a comment regarding the `nproc` dependency, we are good to merge. Thanks!

navsan: Why remove the `-j` flag? 

`nproc` is a coreutil. If it's not there, neither is `tee`. Are you sure you want to add these things to the dependency list?

zuyu: As explained before, users could set `$MAKEFLAGS = -j nproc` to enable concurrent jobs if needed.

In Mac OS, `tee` exists, but not `nproc`.

navsan: OK, I've removed all references to `nproc`. I also removed the Copyright notice, keeping only the license.

navsan:
Thanks Zuyu!

zuyu:
@navsan FYI, please edit the squashed commit messages when doing squashing. Thanks!

PR #24, opened 2016-06-10
Configure libtcmalloc_minimal based on shared libs setting

hbdeshmukh:
- Added a switch to change the way we build libtcmalloc_minimal i.e. either a static library or shared library.
- The build works successfully when tcmalloc is enabled (which is the default setting).

Thanks @navsan for debugging the issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -500,9 +500,13 @@ if(USE_TCMALLOC)
     #   as some generators, e.g. Ninja, may need it to build properly
     # BUILD_BYPRODUCTS <INSTALL_DIR>/lib/libtcmalloc_minimal.a
   )
-  # Static libtcmalloc_minimal.a
-  add_library(libtcmalloc_minimal STATIC IMPORTED)
-  set_property(TARGET libtcmalloc_minimal PROPERTY IMPORTED_LOCATION ${CMAKE_CURRENT_BINARY_DIR}/third_party/gperftools/lib/libtcmalloc_minimal.a)
+  if (BUILD_SHARED_LIBS)
+    add_library(libtcmalloc_minimal SHARED IMPORTED)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -500,9 +500,13 @@ if(USE_TCMALLOC)
...
   )
-  # Static libtcmalloc_minimal.a
-  add_library(libtcmalloc_minimal STATIC IMPORTED)
-  set_property(TARGET libtcmalloc_minimal PROPERTY IMPORTED_LOCATION ${CMAKE_CURRENT_BINARY_DIR}/third_party/gperftools/lib/libtcmalloc_minimal.a)
+  if (BUILD_SHARED_LIBS)
+    add_library(libtcmalloc_minimal SHARED IMPORTED)
+    set_property(TARGET libtcmalloc_minimal PROPERTY IMPORTED_LOCATION ${CMAKE_CURRENT_BINARY_DIR}/third_party/gperftools/lib/libtcmalloc_minimal.so)
+  else()
+    add_library(libtcmalloc_minimal STATIC IMPORTED)
+    set_property(TARGET libtcmalloc_minimal PROPERTY IMPORTED_LOCATION ${CMAKE_CURRENT_BINARY_DIR}/third_party/gperftools/lib/libtcmalloc_minimal.a)
```
zuyu: It seems that we could move this line out of the `if-else` statement.

navsan: Not really, the file name is different based on whether it's a shared or static library.

zuyu: LGTM!

Oh, I missed the last part: `.so` vs `.a`.

zuyu:
@hbdeshmukh I've already merged it to master, but for some reasons, it does not close by asfgit.

Could you please close it? Thanks!

hbdeshmukh:
You mean close the PR right?

PR #25, opened 2016-06-11
DO NOT MERGE: Move createNewBlock() out of the critical section

navsan:
Move createNewBlock() out of the critical section in BlockPoolInsertDestination::getBlockForInsertion.

This significantly improves performance in queries that write a lot of output. 

[Waiting for confirmation from @hbdeshmukh  and @zuyu that this doesn't break anything.]

### DIFF
--- a/storage/InsertDestination.cpp
+++ b/storage/InsertDestination.cpp
@@ -339,21 +339,20 @@ void BlockPoolInsertDestination::getPartiallyFilledBlocks(std::vector<MutableBlo
 }
 
 MutableBlockReference BlockPoolInsertDestination::getBlockForInsertion() {
-  SpinMutexLock lock(mutex_);
-  if (available_block_refs_.empty()) {
-    if (available_block_ids_.empty()) {
-      return createNewBlock();
-    } else {
+  {
+    SpinMutexLock lock(mutex_);
+    if (!available_block_refs_.empty()) {
+      MutableBlockReference retval = std::move(available_block_refs_.back());
+      available_block_refs_.pop_back();
+      return retval;
+    } else if (!available_block_ids_.empty()) {
       const block_id id = available_block_ids_.back();
       available_block_ids_.pop_back();
       MutableBlockReference retval = storage_manager_->getBlockMutable(id, relation_);
       return retval;
     }
-  } else {
-    MutableBlockReference retval = std::move(available_block_refs_.back());
-    available_block_refs_.pop_back();
-    return retval;
   }
+  return createNewBlock();
 }
 
 void BlockPoolInsertDestination::returnBlock(MutableBlockReference &&block, const bool full) {


storage/InsertDestination.cpp
```diff
@@ -339,21 +339,21 @@ void BlockPoolInsertDestination::getPartiallyFilledBlocks(std::vector<MutableBlo
...
-      return createNewBlock();
-    } else {
+  {
+    SpinMutexLock lock(mutex_);
+    if (!available_block_refs_.empty()) {
+      MutableBlockReference retval = std::move(available_block_refs_.back());
+      available_block_refs_.pop_back();
+      return retval;
+    }
+    else if (!available_block_ids_.empty()) {
```
zuyu: Code style: https://google.github.io/styleguide/cppguide.html#Conditionals

```
  if (...) {
  } else if (...) {
  }
```

zuyu:
@navsan This is great, and you may want to fix the same method of other `InsertDestination` implementations.

navsan:
Which ones? And should I do that in the same PR?

zuyu:
Just search the same function name in the same file.

navsan:
I mean, should I do it for all of them, or just some of the more commonly used ones? And can I be sure that createNewBlock() doesn’t need to be in the critical section for all of them? Do we currently use all of them in our code?

Should it be in the same PR or leave that as an issue for later? 

Thanks Zuyu!

zuyu:
`AlwaysCreateBlockInsertDestination::getBlockForInsertion` and
`PartitionAwareInsertDestination::getBlockForInsertionInPartition`.

zuyu:
Fix all in this PR. Thanks!

zuyu:
@navsan I thought you had confirmed that we don't need the critical section when creating a block. OK, let me check it.

navsan:
Harshad confirmed that it doesn’t break anything for BlockPoolInsertDestination. I’m not sure about the others.

storage/InsertDestination.cpp
navsan: Fixed.

storage/InsertDestination.cpp
```diff
@@ -339,21 +339,20 @@ void BlockPoolInsertDestination::getPartiallyFilledBlocks(std::vector<MutableBlo
...
       available_block_ids_.pop_back();
       MutableBlockReference retval = storage_manager_->getBlockMutable(id, relation_);
       return retval;
     }
-  } else {
-    MutableBlockReference retval = std::move(available_block_refs_.back());
-    available_block_refs_.pop_back();
-    return retval;
-  }
+  } // release mutex_ lock
```
zuyu: For this comment, we would need one more whitespace indentation.

```
  ++i;  // Comment.
```

On the other hand, we actually don't need this comment, as it is obvious that the scoped lock automatically releases. Thanks.

navsan: Fixed

zuyu:
@navsan We need to rebase again. The current status that includes changes in 48 files is wired...

navsan:
Wow! This is really going to get some getting used to.

pateljm:
@navsan This does look good to me. The logic makes sense as the mutex there is protecting `available_block_refs_` and `available_block_ids_`, so seems ok to move the new memory allocation outside. One request is to write comments in that function to make the intent clear -- this way if there are issues in the future it will be easier to figure things out. A unit test would be awesome too (as you suggested to me earlier today). 

Thanks a lot for doing this!  

@craig-chasseur -- any comments?

zuyu:
@navsan There are other two places need to change, as a part of the same performance issue.

navsan:
I made all the changes, per @zuyu's comments. But, in a strange turn of events, performance actually worsened when I moved `createNewBlock()` out of the critical section. We didn't get anywhere in investigating why that might be the case. I'm closing this PR for now and leaving it to future work to reopen it, figure out what's going on, and then merge it.

PR #28, opened 2016-06-12
Add option to enable Google Profiler.

navsan:
[Thanks @hbdeshmukh for the pointers on how to do this.]

This PR makes profiling pretty easy. Steps to take: 
* Install dependencies: Google Perftools must be installed already. This PR doesn't use the third_party perftools. 
* Build Quickstep with the `ENABLE_GOOGLE_PROFILER` flag turned on. You probably want to use Build Type `RelWithDebInfo`. 
* When running the CLI shell, pass the profiler_file_name flag to start profiling, and dump profile output to the file name specified. Note that the profiler only starts sampling from the second run of the query, so that it sees warm buffers/caches. Also, the profiler stops collecting samples after the last query run, before the teardown process begins. So we won't see `munmap` in the profile. 
* Alternatively, set the environment variable `CPUPROFILE` to turn on profiling for the entire run (including first run, preloading, foreman and worker setup, etc.).
* Use `google-pprof --dot profile.prof > profile.dot` or something similar to get a usable output. 

Tested with flag turned on and off, with `RelWithDebInfo`, with Clang on CloudLab machine. Dependencies were pre-installed using `sudo apt-get install -y libgoogle-perftools-dev google-perftools`.


cli/QuickstepCli.cpp
```diff
@@ -440,6 +462,13 @@ int main(int argc, char* argv[]) {
     }
   }
 
+#ifdef QUICKSTEP_ENABLE_GOOGLE_PROFILER
+  if (!query_num && !quickstep::FLAGS_profile_file_name.empty()) {
```
zuyu: Look like a bug to me: `query_num` would be `non-zero` if we start to profile.

cli/QuickstepCli.cpp
```diff
@@ -430,6 +445,13 @@ int main(int argc, char* argv[]) {
         reset_parser = true;
         break;
       }
+#ifdef QUICKSTEP_ENABLE_GOOGLE_PROFILER
+      // Profile only if profile_file_name flag is set
+      if (!query_num && !quickstep::FLAGS_profile_file_name.empty()) {
```
zuyu: Question: Do we try to use a singleton pattern here?

cli/QuickstepCli.cpp
navsan: You're right. I'll fix. The profiler produces an output even if you don't explicitly stop the profiler, which is why I didn't catch it. Good catch!

cli/QuickstepCli.cpp
navsan: I'm not sure I understand.

cli/QuickstepCli.cpp
navsan: Fixed now.

cli/QuickstepCli.cpp
zuyu: This is related to the bug we fixed above.

I mean, we should have a bool variable to indicate whether we start profiling, and use that variable to stop profiling, instead of https://github.com/apache/incubator-quickstep/pull/28/files#diff-451b38b1d63242a7d708a368d3c79e21R466.

```
  bool use_profiler = false;
  if (!query_num && !quickstep::FLAGS_profile_file_name.empty()) {
    use_profiler = true;
    ProfilerStart(...);
  }

  if (use_profiler) {
    ProfilerStop();
    ProfilerFlush();
  }
```

cli/QuickstepCli.cpp
```diff
@@ -335,6 +347,9 @@ int main(int argc, char* argv[]) {
   std::unique_ptr<SqlParserWrapper> parser_wrapper(new SqlParserWrapper());
   std::chrono::time_point<std::chrono::steady_clock> start, end;
 
+#ifdef QUICKSTEP_ENABLE_GOOGLE_PROFILER
+  unsigned int query_num = 0;
```
zuyu: We could remove it then.

cli/QuickstepCli.cpp
zuyu: @navsan Ping again this issue. I did not see we use a bool variable here. Do we go with any alternative options?

cli/QuickstepCli.cpp
navsan: See the latest version. There's a bool variable declared here instead. We need the declaration to be outside the for loop, so there'll still be a code change here.

cli/QuickstepCli.cpp
navsan: @zuyu, Sorry, was AFK. The latest version does something similar.

cli/QuickstepCli.cpp
```diff
@@ -430,6 +445,13 @@ int main(int argc, char* argv[]) {
         reset_parser = true;
         break;
       }
+#ifdef QUICKSTEP_ENABLE_GOOGLE_PROFILER
+      // Profile only if profile_file_name flag is set
+      if (!started_profiling && !quickstep::FLAGS_profile_file_name.empty()) {
```
zuyu: Now we are using the singleton design pattern.

I still have a question: why we start profiling after one query finishes execution, instead of before any query starts? Thanks!

navsan: Good question. It’s because, unless you’ve preloaded the buffer pool (not always a good idea), the first run of the query results in disk I/O and other overhead that significantly skews the profiling results. It’s the same reason we don’t include the first run time in our benchmarking: when profiling query execution, it makes more sense to get numbers using a warm buffer pool and warm caches. This is not *always* the right thing to do: it’s obviously wrong for profiling the TextScan operator. In those cases, you might want to put in your own Profiler probes (just follow the start/stop pattern used in this PR) or just run quickstep with the CPUPROFILE environment variable set (as per gperftools documentation) to get the full profile for the entire execution. 

I’ve provided the explanation in the flag definition, CMake comments as well as in the PR header comment.

cli/QuickstepCli.cpp
```diff
@@ -150,6 +154,14 @@ DEFINE_bool(initialize_db, false, "If true, initialize a database.");
 DEFINE_bool(print_query, false,
             "Print each input query statement. This is useful when running a "
             "large number of queries in a batch.");
+DEFINE_string(profile_file_name, "",
```
hbdeshmukh: One thing which we can do is generate a default profile name if none is specified. The default profile name can be something like "quickstep-profile-<timestamp>.prof", so that name collisions are avoided. That way, we can get rid of the check ``!quickstep::FLAGS_profile_file_name.empty()`` in Line 450. It's ok if we don't do it right away, in which case you can leave a TODO note.

cli/QuickstepCli.cpp
zuyu: Thanks for the reply.

Then, we should also put the explanation as a comment in the file, and we'll be good to merge. Thanks!

cli/QuickstepCli.cpp
navsan: Actually, I’d argue that this is a *feature*. We can build with profiling included, but enable it only when we want to (rather than doing so for every run).

hbdeshmukh: Alright. So, if I get it right, we need to provide three parameters, BUILD_TYPE as RelWithDebInfo, ENABLE_GOOGLE_PROFILER as ON and a profile_file_name string for getting the profiling output. There are two options, which if achievable, can make things a little bit simpler. 
1. When someone says ENABLE_GOOGLE_PROFILER is ON, we automatically change the BUILD_TYPE to RelWithDebInfo OR
2. Create a new BUILD_TYPE as "Profile", which internally takes care of enabling the profiling, ensuring that the necessary debug symbols are present in the code and optionally choosing a default profiling output filename, if one not specified. Once the quickstep process terminates, we can just print a message displaying the filename of the profiling output.

navsan: Thanks @hbdeshmukh for those suggestions. I'm not sure I completely agree with them, but before we discuss them, are you suggesting that we make these changes in the current PR or put it off to another one? If the latter, let's merge and discuss this on another thread (maybe a JIRA issue?).

hbdeshmukh: We can merge this one now and discuss the comments in a future PR.

cli/QuickstepCli.cpp
navsan: Added the explanation in the the QuickstepCli.cpp file. Are we good to merge? 
Thanks @zuyu, @hbdeshmukh for the thorough comments!

hbdeshmukh:
LGTM. We can merge after the conflicts are resolved.

navsan:
Thanks. I'll resolve and resubmit. I seem to have messed something up in my local branch.

navsan:
OK, it should be ready now, I think.

cli/QuickstepCli.cpp
zuyu: @navsan I did not see the detailed explanation beyond `Profile only if profile_file_name flag is set`. Did I miss something?

navsan: It’s at the top, below the definition of the flag. It’s more useful there, since it’s really a comment about how to use the flag. But it’s too long to be part of the flag description itself.

zuyu:
LGTM.

zuyu:
Merged!

navsan:
Thanks Zuyu!

PR #29, opened 2016-06-13
Fix bug: specify log file name correctly

navsan:
I introduced a bug during one of the many merges!


pateljm:
LGTM. Merging. Thanks @navsan!

PR #30, opened 2016-06-14
Fix conditional per-target flags for lexer

craig-chasseur:
set_target_properties() overwrites per-target COMPILE_FLAGS rather than replacing them. Also, compiler flags should be checked before enabling them.


pateljm:
Merged.

PR #32, opened 2016-06-14
QUICKSTEP-21 Measure execution time of WorkOrders

hbdeshmukh:
This PR includes the following changes: 

- Measure work order execution time
- Split the WorkOrderCompletion proto message in two: Normal and Rebuild work orders.
- Add execution time field in the both types of WorkOrderComplete proto messages.
- Include the recorded time in both kind of WorkOrderComplete proto messages that are sent back to Foreman.

Also note that, the CPU overhead for measuring this time is pretty low. In terms of memory, we pay a cost of an additional 8 bytes in each completion message sent from each worker to the Foreman. 

This feature can be quite helpful for performance monitoring. We can take a look at individual work order execution time of different operators. I am planning to create another PR which can produce reports consisting of average/max/min/variance of execution times grouped by different operators.

### DIFF
--- a/query_execution/QueryExecutionMessages.proto
+++ b/query_execution/QueryExecutionMessages.proto
@@ -20,14 +20,33 @@ package quickstep.serialization;
 message EmptyMessage {
 }
 
-// Used for both Normal WorkOrders and RebuildWorkOrders.
-// NOTE(zuyu): we might need to seperate the completion messages to contain
-// run-time information for Foreman to make better decisions on scheduling
-// WorkOrders.
-message WorkOrderCompletionMessage {
+// Note: There are different types of completion messages for normal work orders
+// rebuild work orders. This can be potentially helpful when we want to collect
+// different statistics for executing different types of work orders. 
+// e.g. In select normal work order completion message, we could be interested 
+// in the selectivity of the block whose work order got execute. In rebuild work 
+// order completion message, we may be interested in adding the compression 
+// ratio or dictionary size of the rebuilt block.
+
+// TODO(harshad) : If there are different fields in the two message types below, 
+// create a base message class called WorkOrderCompletionMessage and make the
+// two classes below extend the base class. All the common fields in both the
+// classes can be moved to the base class. 
+
+// A message sent upon completion of a normal (not rebuild) WorkOrder execution.
+message NormalWorkOrderCompletionMessage {
   required uint64 operator_index = 1;
   required uint64 worker_thread_index = 2;
   required uint64 query_id = 3;
+  optional uint64 execution_time_in_microseconds = 4;
+}
+
+// A message sent upon completion of a rebuild WorkOrder execution.
+message RebuildWorkOrderCompletionMessage {
+  required uint64 operator_index = 1;
+  required uint64 worker_thread_index = 2;
+  required uint64 query_id = 3;
+  optional uint64 execution_time_in_microseconds = 4;
 }
 
 message CatalogRelationNewBlockMessage {
--- a/query_execution/Worker.cpp
+++ b/query_execution/Worker.cpp
@@ -16,7 +16,9 @@
 
 #include "query_execution/Worker.hpp"
 
+#include <chrono>
 #include <cstddef>
+#include <cstdint>
 #include <cstdlib>
 #include <utility>
 
@@ -36,6 +38,7 @@
 #include "tmb/tagged_message.h"
 
 using std::size_t;
+using std::uint64_t;
 
 using tmb::TaggedMessage;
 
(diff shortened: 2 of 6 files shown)


zuyu:
@hbdeshmukh Since the time field in `WorkOrderCompletion` proto is optional, we still do not need to split into two.

query_execution/QueryExecutionMessages.proto
```diff
@@ -20,11 +20,16 @@ package quickstep.serialization;
...
-// NOTE(zuyu): we might need to seperate the completion messages to contain
-// run-time information for Foreman to make better decisions on scheduling
-// WorkOrders.
-message WorkOrderCompletionMessage {
+// A message sent upon completion of a normal (not rebuild) WorkOrder execution.
+message NormalWorkOrderCompletionMessage {
+  required uint64 operator_index = 1;
+  required uint64 worker_thread_index = 2;
+  required uint64 query_id = 3;  
+  optional uint64 execution_time_us = 4;  // Execution time in micro seconds.
```
zuyu: Prefer to `execution_time_in_microseconds`.

query_execution/Worker.cpp
```diff
@@ -47,13 +49,32 @@ void Worker::run() {
...
+        WorkerMessage message(*static_cast<const WorkerMessage*>(tagged_message.message()));
+        DCHECK(message.getWorkOrder() != nullptr);
+        // Start measuring the execution time.
+        start = std::chrono::steady_clock::now();
+        message.getWorkOrder()->execute();
+        end = std::chrono::steady_clock::now();
+        const std::size_t query_id_for_workorder =
+            message.getWorkOrder()->getQueryID();
+        delete message.getWorkOrder();
+        const uint64_t time_us =
```
zuyu: Either use `std::uint64_t`, or add `using std::uint64_t;` before the namespace.

query_execution/QueryExecutionMessages.proto
hbdeshmukh: Changed.

query_execution/Worker.cpp
hbdeshmukh: Changed. Thanks for pointing out.

hbdeshmukh:
Hi @zuyu 

You are right that at this point we can keep the two proto message classes combined. However in the future, in order to add more information to the normal WorkOrder completion message (e.g. resources consumed during the WorkOrder execution, number of tuples touched etc.) I think it makes more sense to split the two classes early enough.

zuyu:
Even for the info you mentioned, `RebuildWorkOrder` could also have these stored in `proto`.

hbdeshmukh:
We are interested in such information only for normal work orders, not for rebuild work orders. So why should we add unnecessary fields in the rebuild completion message? 

Is there any other reason behind your resistance for the change, which I am missing?

pateljm:
Good discussion guys. If I could weigh in, I think keeping the two proto classes makes sense. I'd like to close this later on today if the discussion here has converged. Thanks!

pateljm:
@hbdeshmukh Can you rebase this again? I can then merge. Thanks!

zuyu:
@pateljm Separating into two proto messages is overkill to me. In fact, the most part in this PR is to duplicate the logics to process two identical proto messages, except for `normal` and `rebuild`. Note that doing such measurement only needs about 10 lines of code changes.

On the other hand, we do differentiate two messages using two distinct `TMB` message types.

pateljm:
@zuyu @hbdeshmukh Ok -- looking at the code, I do think that having two protos make sense. Down the line `RebuildWorkOrderCompletionMessage` may need to report back compression ratio and/or block under-fill stats (for compressed column stores). These won't make sense to include in the proto definition for `NormalWorkOrderCompletionMessage`. So, let's keep these two protos separate. We could do a union, but that is ugly and confusing code.

@zuyu has a good point that (if I read part of his concern correctly) that we do need to record and report back `execution_time_in_microseconds` even for `RebuildWorkOrderCompletionMessage`. So let's do that. 

Also let's make a note in the comments around the proto definition about why we have two protos (i.e. note that in the future `RebuildWorkOrderCompletionMessage` will report back additional stats), so that the reader is not confused.

I recommend @hbdeshmukh decide if he wants to fix this issue in this PR, or open a separate PR relatively soon. If the latter is chosen, please make a note here in the discussion thread for this PR.

In general, I'd recommend in situations like this to have a healthy discussion. Then, if we are in a gray area, let us document the concern, and give discretion to the person opening the PR.

I want to thank @zuyu for the diligent review -- without such careful review, the code base could quickly become unmanageable. So keep it coming, and don't hold back in the future.

hbdeshmukh:
Hi @pateljm @zuyu 

I can get the execution time for RebuildWorkOrder in the same PR. Thanks for your comments.

zuyu:
@pateljm I agree that if we would go with the split option, we do w/ extensions as following.

```
message WorkOrderCompletionMessage {
  ...
}

message NormalWorkOrderCompletionMessage {
  extend WorkOrderCompletionMessage {
    ...
  }
}

message RebuildWorkOrderCompletionMessage {
  extend WorkOrderCompletionMessage {
    ...
  }
}
```

hbdeshmukh:
Hi @pateljm @zuyu 

I have added the execution time field in rebuild work order completion proto message. There are detailed comments about the rationale for splitting the original class in two types. I have also added a TODO regarding @zuyu's suggestion on message classes extension, which we can address in a future PR. That PR should leave the Foreman and PolicyEnforcer classes untouched.

pateljm:
Thanks @hbdeshmukh. Merging now.

zuyu:
@pateljm Please also delete the branch after merging, as a part of the workflow in the wiki. Thanks!

PR #33, opened 2016-06-15
BugFix: Update NumQueuedWorkOrders to fix scheduling

navsan:
Quickstep scheduling is currently broken (since PR #14). The foreman only schedules work for one worker, leaving all other workers idle. This PR fixes that bug. 

The foreman maintains the number of queued work orders for each worker in the WorkerDirectory. This state was not being incremented when WorkOrders are dispatched, and not being decremented when WorkOrders are completed. The search for LeastLoadedWorker would therefore always pick the first worker, resulting in serial execution of the entire query. 

In this PR, I have incremented the number of queued work orders in the Foreman when it dispatches messages. 

When a WorkOrderCompletion message arrives, this number must be decremented. However, the worker's thread ID is not available to the Foreman, since PR #14 moved the message deserialization into the PolicyEnforcer. So, in this PR, I've added a pointer to the WorkerDirectory in the PolicyEnforcer. The PolicyEnforcer decrements the number of queued work orders while processing WorkOrderCompletion messages. 

The WorkerDirectory is not thread-safe, so the decrement should only be done in the Foreman thread. Since the PolicyEnforcer is part of the Foreman thread, this change should be fine. 

Tested on CloudLab machine  (40 workers) with a few example queries. 

[A big thanks to @rogersjeffreyl for helping me debug this!]

### DIFF
--- a/query_execution/PolicyEnforcer.hpp
+++ b/query_execution/PolicyEnforcer.hpp
@@ -40,6 +40,7 @@ namespace quickstep {
 class CatalogDatabaseLite;
 class QueryHandle;
 class StorageManager;
+class WorkerDirectory;
 
 /**
  * @brief A class that ensures that a high level policy is maintained
@@ -60,11 +61,13 @@ class PolicyEnforcer {
                  const std::size_t num_numa_nodes,
                  CatalogDatabaseLite *catalog_database,
                  StorageManager *storage_manager,
+                 WorkerDirectory *worker_directory,
                  tmb::MessageBus *bus)
       : foreman_client_id_(foreman_client_id),
         num_numa_nodes_(num_numa_nodes),
         catalog_database_(catalog_database),
         storage_manager_(storage_manager),
+        worker_directory_(worker_directory),
         bus_(bus) {}
 
   /**
@@ -148,6 +151,7 @@ class PolicyEnforcer {
 
   CatalogDatabaseLite *catalog_database_;
   StorageManager *storage_manager_;
+  WorkerDirectory *worker_directory_;
 
   tmb::MessageBus *bus_;
 
--- a/query_execution/CMakeLists.txt
+++ b/query_execution/CMakeLists.txt
@@ -93,6 +93,7 @@ target_link_libraries(quickstep_queryexecution_PolicyEnforcer
                       quickstep_queryexecution_QueryExecutionMessages_proto
                       quickstep_queryexecution_QueryExecutionTypedefs
                       quickstep_queryexecution_QueryManager
+                      quickstep_queryexecution_WorkerDirectory
                       quickstep_queryexecution_WorkerMessage
                       quickstep_queryoptimizer_QueryHandle
                       quickstep_relationaloperators_WorkOrder
--- a/query_execution/Foreman.cpp
+++ b/query_execution/Foreman.cpp
@@ -89,6 +89,7 @@ Foreman::Foreman(const tmb::client_id main_thread_client_id,
       num_numa_nodes,
       catalog_database_,
       storage_manager_,
+      worker_directory_,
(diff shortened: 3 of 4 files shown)


hbdeshmukh:
Thanks @navsan and @rogersjeffreyl for fixing this. This was an oversight on my part. I checked the correctness of queries for PR #14, but I should have also tested the performance, which would have revealed the bug. 

This looks good to me. I can merge the PR once travis signals green.

navsan:
Awesome, thanks!

query_execution/PolicyEnforcer.hpp
```diff
@@ -60,12 +61,14 @@ class PolicyEnforcer {
                  const std::size_t num_numa_nodes,
                  CatalogDatabaseLite *catalog_database,
                  StorageManager *storage_manager,
-                 tmb::MessageBus *bus)
+                 tmb::MessageBus *bus,
+                 WorkerDirectory *worker_directory)
```
zuyu: This is optional, but I would suggest to move `worker_directory` before `bus` under the idea that puts all third party stuff last.

navsan:
@hbdeshmukh I think it's rather bug-prone to maintain queue state in the Foreman. Why not just look it up in the TMB itself? I noticed that you made the same comment above the WorkerDirectory class.

query_execution/PolicyEnforcer.hpp
navsan: Changed the order. BTW, this is not a convention that is followed elsewhere in the file.

navsan:
Travis failures were due to CMake validation. Build and tests succeeded. I've fixed the validation failure as well now.

hbdeshmukh:
LGTM, merging.

hbdeshmukh:
That's true. I don't think it is too bug-prone. We just need to just increment and decrement the count at a couple of places (which I missed, admittedly). 

What TMB gives us is perhaps more accurate snapshot of the queue lengths of each worker compared to the queue lengths as viewed by Foreman. However the scheduling decisions themselves are much faster than a typical WorkOrder execution, so the lag (and corresponding difference in queue lengths) may not be too big an issue. 

On the other hand, I am not sure of the performance implications of checking the queue lengths through TMB repeatedly, e.g. to find the least loaded worker before every work order execution. 

We may not be able to get rid of WorkerDirectory completely either. Consider a case when you want to find out the least loaded worker on a specific NUMA socket. TMB doesn't recognize this case and I don't want to modify TMB code either. In such a case we need a level of indirection that gives us a list of workers on that NUMA socket.

pateljm:
@navsan Nice job!! Thanks @rogersjeffreyl for helping. 

@hbdeshmukh Re. your question about checking the TMB, I think you only need to check the TMB periodically (i.e. for a new back of work orders). This may not be too bad even in the distributed case, but certainly checking for each queue length in the TMB for each work order dispatch is too heavy-weight. I see a nice opportunity for refactoring the code here, but we can do that in a future PR. Would be great to make a note of this refactoring task in JIRA.

pateljm:
This also points out that we should have a better CI in place for performance testing. We periodically run benchmark tests, but that is not in a CI pipeline. Travis is not ideal for this (no guarantees on the underlying VM that is created), but there are tools like [blitz.io](https://www.blitz.io/)  and [blazemeter](https://www.blazemeter.com/) that could work. Added a [JIRA ticket](https://issues.apache.org/jira/browse/QUICKSTEP-23) to make a note of this.

zuyu:
@hbdeshmukh is right about `WorkerDirectory`, and enforcing `NUMA-aware` scheduling is the main reason to introduce this class.

In the distributed version, `Foreman` needs to query the load of `Shiftboss`.

PR #34, opened 2016-06-15
Bug fixed in \analyze command and reuse code.

hbdeshmukh:
The ``\analyze`` command issues SQL queries. Due to a recent change in the execution engine (PR #14), there was a bug in issuing the queries, which is fixed in this branch. The issue is described below:

The workflow to issue queries is as follows:

1. Create a TMB message consisting of the query. 
2. Send the TMB message from main thread to the Foreman thread.
3. Foreman executes the query.
4. Foreman sends a completion message back to the main thread.
5. Main thread receives the completion message and proceeds. 

In the relevant function where ``\analyze`` queries are issued, we were not capturing the response in step 5, which gave a segmentation fault sometimes. 

This branch also puts a common code to receive feedback from Foreman upon query completion in a function, which gets used in multiple places. There are some minor inclusion fixes in the touched files.

### DIFF
--- a/cli/CommandExecutor.cpp
+++ b/cli/CommandExecutor.cpp
@@ -220,6 +220,8 @@ inline TypedValue executeQueryForSingleResult(
   QueryExecutionUtil::ConstructAndSendAdmitRequestMessage(
       main_thread_client_id, foreman_client_id, query_handle.get(), bus);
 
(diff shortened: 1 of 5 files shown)


hbdeshmukh:
Assigning @jianqiao

jianqiao:
LGTM! Merging.

PR #36, opened 2016-06-16
QUICKSTEP-24 Report generation support for work order execution statistics

hbdeshmukh:
This PR provides an initial version of the support required to report individual work order execution performance. Some highlights are described below: 

- A flag to enable work order profiling report generation.
- At the end of each query, a report is generated which includes worker
  ID, its NUMA socket, the query ID, the operator that produced the WorkOrder and the
  execution time in microseconds.
- The output is printed on stdout in CSV format as of now.

As this is a rudimentary support for the functionality, there is a lot of future work in this regards, which includes printing of CPU core information, printing operator name, allowing user to specify a file where the output can be written, extend the support for rebuild WorkOrder statistics etc.

Edit: Added query ID.

### DIFF
--- a/query_execution/Foreman.cpp
+++ b/query_execution/Foreman.cpp
@@ -18,7 +18,9 @@
 #include "query_execution/Foreman.hpp"
 
 #include <cstddef>
+#include <cstdio>
 #include <memory>
+#include <tuple>
 #include <utility>
 #include <vector>
 
@@ -54,7 +56,8 @@ Foreman::Foreman(const tmb::client_id main_thread_client_id,
                  CatalogDatabaseLite *catalog_database,
                  StorageManager *storage_manager,
                  const int cpu_id,
-                 const size_t num_numa_nodes)
+                 const size_t num_numa_nodes,
+                 const bool profile_individual_workorders)
     : ForemanLite(bus, cpu_id),
       main_thread_client_id_(main_thread_client_id),
       worker_directory_(DCHECK_NOTNULL(worker_directory)),
@@ -90,7 +93,8 @@ Foreman::Foreman(const tmb::client_id main_thread_client_id,
       catalog_database_,
       storage_manager_,
       worker_directory_,
-      bus_));
+      bus_,
+      profile_individual_workorders));
 }
 
 void Foreman::run() {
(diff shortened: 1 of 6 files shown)


pateljm:
@hbdeshmukh This is a neat feature. LGTM. Will merge once Travis comes back green. (BTW, we don't need to edit the copyright files anymore, but don't worry about it in this PR.)

CMakeLists.txt
```diff
@@ -185,6 +185,11 @@ if (ENABLE_GOOGLE_PROFILER)
   set(LIBS ${LIBS} profiler)
 endif()
 
+# This option enables profiling individual work order performances. If enabled, 
+# at the end of each query execution, we print on stdout in CSV format the 
+# statistics about every work order that was executed. 
+option(ENABLE_WORKORDER_PROFILING "Enable profiling individual WorkOrder execution" OFF)
```
zuyu: What do you think of using `gflags` in the run-time, instead of this compilation time flag?

hbdeshmukh: Hi @zuyu 

Jianqiao and I had a chat about this and we agreed that gflags should be a better option instead of compile time flags. I will make that change.

zuyu: Cool.

query_execution/Foreman.cpp
```diff
@@ -229,4 +229,22 @@ void Foreman::sendWorkerMessage(const size_t worker_thread_index,
...
+      &recorded_times = policy_enforcer_->getProfilingResults();
+  fputs("Worker ID, NUMA Socket, Query ID, Operator ID, Time (microseconds)\n", out);
+  for (auto workorder_entry : recorded_times) {
+    // Note: Index of the "worker thread index" in the tuple is 0.
+    const std::size_t worker_id = std::get<0>(workorder_entry);
+    fprintf(out, "%lu,", worker_id);
+    fprintf(out, "%d,", worker_directory_->getNUMANode(worker_id));
+    fprintf(out, "%lu,", std::get<1>(workorder_entry));
+    fprintf(out, "%lu,", std::get<2>(workorder_entry));
+    fprintf(out, "%lu\n", std::get<3>(workorder_entry));
```
zuyu: How about merging these `fprintf`s into one?

Although this is not in a critical path, we still save tons of system calls.

pateljm: @zuyu makes a good point. Would be ok to do that in a separate PR, if it is too much to do in this one. Thanks @hbdeshmukh for this feature!

hbdeshmukh:
Hi @pateljm 

I have made the requested changes. The API to get the stats is changed a bit. Now the functions ``printWorkOrderProfilingResults()`` and ``getProfilingResults()`` require a query ID to be passed, so that we can print results for queries with specific ID. Typically, in our experiments we ignore the first run of the query, so we can use the profiling results for subsequent runs. 

I have performed basic testing on the command line with more than one queries as input and it worked alright.

pateljm:
@hbdeshmukh Ready to merge, but need to rebase this ahead of the master. Let's close this one first before PR #38.

hbdeshmukh:
Hi @pateljm 

Rebased with master.

PR #38, opened 2016-06-17
QUICKSTEP-25 PRIORITY clause in SQL statements

hbdeshmukh:
- SQL statements (right now only SELECT statements) support "WITH PRIORITY N" clause.
- The priority value N should be a non-zero positive integer, with default as 1.
- The priority value is a part of the QueryHandle class.


hbdeshmukh:
Assigning @rogersjeffreyl (and thanks for the help during the development).

parser/preprocessed/SqlLexer_gen.cpp
```diff
@@ -8,8 +8,8 @@
 
 #define FLEX_SCANNER
 #define YY_FLEX_MAJOR_VERSION 2
-#define YY_FLEX_MINOR_VERSION 6
-#define YY_FLEX_SUBMINOR_VERSION 0
+#define YY_FLEX_MINOR_VERSION 5
```
hbdeshmukh: Do I need to build the code on a machine with FLEX version as 2.6.0 and check the preprocessed files in?

parser/ParseStatement.hpp
```diff
@@ -83,6 +84,10 @@ class ParseStatement : public ParseTreeNode {
    **/
   virtual StatementType getStatementType() const = 0;
 
+  virtual const std::int64_t getPriority() const {
```
rogersjeffreyl: Please add comments

parser/preprocessed/SqlLexer_gen.cpp
```diff
@@ -1300,11 +1280,16 @@ extern int quickstep_yylex \
  */
 YY_DECL
 {
-	yy_state_type yy_current_state;
-	char *yy_cp, *yy_bp;
-	int yy_act;
+	register yy_state_type yy_current_state;
```
rogersjeffreyl: Please revert all the variables that have been modified to  register  type

parser/preprocessed/SqlLexer_gen.cpp
rogersjeffreyl: The master branch lists 2.5.38 as  the minimum requirement for the FLEX version. So any version > than that should  be fine. However, there is a TODO  
 # TODO(chasseur): This might not be the exact version to test for.
  if (FLEX_VERSION VERSION_GREATER 2.5.38)
We might need to get this fixed.

rogersjeffreyl: @pateljm Right now the travis in apache runs on a older version of FLEX. Hence it uses the files present in parser/preprocessed by default.  Any one who adds new tokens/rules to the parser has to run parser/genfiles.sh  and check the generated files in  for the changes to work.  We should be making sure that this is a part of the documentation for the parser workflow.

hbdeshmukh: Sure. My machine has Flex 2.5.37. I can work with you to generate the pre-processed files for version 2.5.38+. 

How about the requirements for Bison version?

parser/preprocessed/SqlLexer_gen.cpp
hbdeshmukh: Will 2.5.38+ produce a cpp file without the register type? Or I need to modify the file manually (even with a higher Flex version) by removing register keyword?

hbdeshmukh: Also, is there any compilation issue with having register type?

parser/preprocessed/SqlLexer_gen.cpp
pateljm: Good point @rogersjeffreyl on noting it in the developer workflow. Done! Changes are reflected in https://cwiki.apache.org/confluence/display/QUICKSTEP/Workflow+For+Committer

pateljm:
@hbdeshmukh LGTM. Ready to merge once you rebase as the master has moved with the last PR that closed. Sorry!

parser/preprocessed/SqlLexer_gen.cpp
rogersjeffreyl: @hbdeshmukh 
The register variables are generated in my my machine as well ans I get a compilation error. However it seems to have worked for you. Not sure if this is due to the version of flex being used.
We can generate the modified   preprocessed files on my machine and  then commit them

hbdeshmukh:
Hi @pateljm @rogersjeffreyl 

I have addressed the review comments. I updated the pre-processed parser files using flex version 2.6.0 (the highest available) version right now that doesn't produce ``register`` keyword in the cpp/hpp files.

rogersjeffreyl:
@hbdeshmukh  Looks good. We can merge it once travis passes.

zuyu:
Next time, please delete the branch after merged.

PR #39, opened 2016-06-23
QUICKSTEP-20: Resolver support for Window Aggregate Function

shixuan-fan:
Now the resolver could form a logical plan for window aggregate function based on the parsed result.

Suppose we have the following query:
SELECT attr0, avg(attr1) OVER
(PARTITION BY attr2
ORDER BY attr3
ROWS BETWEEN 3 PRECEDING AND 3 FOLLOWING)
FROM table1;

The logical plan will look like this:
input_table -> sort on (attr2, attr3) -> window aggregate -> project on (attr0, avg(attr1)).

Two new classes are introduced: quickstep::optimizer::expressions::WindowAggregateFunction and quickstep::optimizer::logical::WindowAggregate:
  -WindowAggregateFunction will keep the function information and window information. 
  -WindowAggregate will store the input logical node and an alias wrapper of WindowAggregateFunction.

In current version, we only support one window aggregate function. When the PhysicalPlanGenerator sees a logical node with type kWindowAggregate, it will throw an error saying that "window aggregate function is not supported".

PR #40, opened 2016-06-27
Minor Change: Use move semantics when constructing optimizer::logical::Sort

shixuan-fan:
Added move constructor in optimizer::logical::Sort.

### DIFF
--- a/query_optimizer/logical/Sort.hpp
+++ b/query_optimizer/logical/Sort.hpp
@@ -113,6 +113,19 @@ class Sort : public Logical {
                             limit));
   }
 
+  static SortPtr Create(
+      const LogicalPtr &input,
+      std::vector<expressions::AttributeReferencePtr> &&sort_attributes,
+      std::vector<bool> &&sort_ascending,
+      std::vector<bool> &&nulls_first_flags,
+      const int limit) {
+    return SortPtr(new Sort(input,
+                            std::move(sort_attributes),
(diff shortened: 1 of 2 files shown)


query_optimizer/logical/Sort.hpp
```diff
@@ -137,6 +150,19 @@ class Sort : public Logical {
...
+       const std::vector<bool> &&sort_ascending,
+       const std::vector<bool> &&nulls_first_flags,
+       const int limit)
+      : input_(input),
+        sort_attributes_(std::move(sort_attributes)),
+        sort_ascending_(std::move(sort_ascending)),
+        nulls_first_flags_(std::move(nulls_first_flags)),
+        limit_(limit) {
+    addChild(input_);
+  }
```
jianqiao: Remove `const` for all rvalue reference (`&&`) arguments, otherwise it is not really doing the move construction.

shixuan-fan: Thanks! Removed all six `const`s :)

zuyu:
The title is misleading, please change to `Use move semantics when constructing optimizer::logical::Sort`, since we are not adding a move constructor, but adding a new constructor w/ some rvalue arguments.

Also, we need another separate PR for `WindowPlan`, as discussed in [here](https://github.com/apache/incubator-quickstep/pull/39#discussion_r68328569).

PR #41, opened 2016-06-27
Remove misleading comment about hash table collector type

navsan:
Consulted with @hbdeshmukh who confirmed that these comments are misleading.

### DIFF
--- a/relational_operators/HashJoinOperator.cpp
+++ b/relational_operators/HashJoinOperator.cpp
@@ -65,21 +65,11 @@ DEFINE_bool(vector_based_joined_tuple_collector, false,
             "If true, use simple vector-based joined tuple collector in "
             "hash join, with a final sort pass to group joined tuple pairs "
             "by inner block. If false, use unordered_map based collector that "
-            "keeps joined pairs grouped by inner block as they are found "
-            "(this latter option has exhibited performance/scaling problems, "
-            "particularly in NUMA contexts).");
+            "keeps joined pairs grouped by inner block as they are found.");
 
 // Functor passed to HashTable::getAllFromValueAccessor() to collect matching
 // tuples from the inner relation. This version stores matching tuple ID pairs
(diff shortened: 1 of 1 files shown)


relational_operators/HashJoinOperator.cpp
```diff
@@ -65,21 +65,11 @@ DEFINE_bool(vector_based_joined_tuple_collector, false,
             "If true, use simple vector-based joined tuple collector in "
             "hash join, with a final sort pass to group joined tuple pairs "
             "by inner block. If false, use unordered_map based collector that "
-            "keeps joined pairs grouped by inner block as they are found "
-            "(this latter option has exhibited performance/scaling problems, "
-            "particularly in NUMA contexts).");
+            "keeps joined pairs grouped by inner block as they are found.");
```
hbdeshmukh: Hi @navsan 

How about changing the flag name to ``map_based_joined_tuple_collector`` and switch the default to true?

hbdeshmukh:
Hi @navsan 

I have one minor comment, otherwise LGTM. Thanks for the cleanup.

relational_operators/HashJoinOperator.cpp
navsan: Sure, I can change that. 
Do we actually need the vector_based collector at all? If it’s unclear whether it will ever outperform map_based, maybe we should just get rid of it?

hbdeshmukh: I don't mind removing it. If I recall correctly, an older profiling done by @pateljm revealed that map collector outperformed the vector collector pretty much always.

navsan: My rudimentary profiling also gave the same results. I’ll wait for others to opine. If there are no objections, I’ll remove it.

navsan:
I'm going to close this PR, and remove the vector_based collector type instead.

PR #43, opened 2016-06-27
Remove unused AlwaysCreateBlockInsertDestination

navsan:
As per [discussion](http://mail-archives.apache.org/mod_mbox/incubator-quickstep-dev/201606.mbox/%3CCAA-fBsovOVutKi%3DfvTpQ8FMW%3DTf8rSSj8H-3JB7QHGXuYFkVyA%40mail.gmail.com%3E) with @adalbertgerald and @hbdeshmukh in the dev mailing list.

### DIFF
--- a/storage/InsertDestination.cpp
+++ b/storage/InsertDestination.cpp
@@ -92,15 +92,6 @@ InsertDestination* InsertDestination::ReconstructFromProto(
   }
 
   switch (proto.insert_destination_type()) {
-    case serialization::InsertDestinationType::ALWAYS_CREATE_BLOCK: {
-      return new AlwaysCreateBlockInsertDestination(relation,
-                                                    layout,
-                                                    storage_manager,
-                                                    proto.relational_op_index(),
-                                                    query_id,
-                                                    scheduler_client_id,
-                                                    bus);
-    }
     case serialization::InsertDestinationType::BLOCK_POOL: {
       vector<block_id> blocks;
       for (int i = 0; i < proto.ExtensionSize(serialization::BlockPoolInsertDestination::blocks); ++i) {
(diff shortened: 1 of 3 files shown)


hbdeshmukh:
Hi @navsan 

We need to update the storage/InsertDestination.proto to reflect the removal of this derived class.

hbdeshmukh:
Do we need the optional boolean parameter ``always_mark_full = false`` in the ``bulkInsert*`` methods now? I don't see it being set to ``true`` anywhere.

navsan:
Hi @hbdeshmukh, I've made the corresponding change to the .proto file. 

On the `always_mark_full` flag: I agree that it doesn't seem to be used anywhere. But is it directly related to the `AlwaysCreateBlockInsertDestination`? I didn't see any comments explaining why this flag was added, and the flag is part of the `InsertDestination` interface. Just to be safe, I've made the change in a separate commit, rather than squashing it.

zuyu:
@navsan Hold, we may need it for the long-standing streaming queries.

navsan:
Why? Why not just return blocks to the storage manager marking them as full? There should be negligible overhead in that case (2 branch instructions that are always predicted false; once for every block/work order).

navsan:
@hbdeshmukh We were wrong: `always_mark_full` is indeed being used [in the sort function](https://github.com/apache/incubator-quickstep/blob/master/storage/StorageBlock.cpp#L940). I'm not sure why it's necessary there, but I'd rather not go change that now. I'm reverting that commit and only submitting the removal of the AlwaysCreateBlockInsertDestination (first commit).

navsan:
@zuyu Any objections to my proposal [above](https://github.com/apache/incubator-quickstep/pull/43#issuecomment-231207534)? Can I go ahead with the submission?

zuyu:
@navsan No, we have to close this PR w/o merging, because we still need it as a workaround of lacking the concurrent control over multiple update queries.

navsan:
i’m sorry, I don’t quite follow that. Would you mind elaborating?

pateljm:
@navsan: Can we resolve this PR? Perhaps, pull back on this? Thanks!

PR #45, opened 2016-06-28
QUICKSTEP-20: PhysicalGenerator support for Window Aggregation

shixuan-fan:
- The PhysicalGenerator could now generate a physical plan for window aggregate function using OneToOne strategy. 
- The ExecutionGenerator will catch window aggregate function and throw an exception.

### DIFF
--- a/query_optimizer/ExecutionGenerator.cpp
+++ b/query_optimizer/ExecutionGenerator.cpp
@@ -83,6 +83,7 @@
 #include "query_optimizer/physical/TableReference.hpp"
 #include "query_optimizer/physical/TopLevelPlan.hpp"
 #include "query_optimizer/physical/UpdateTable.hpp"
+#include "query_optimizer/physical/WindowAggregate.hpp"
 #include "relational_operators/AggregationOperator.hpp"
 #include "relational_operators/BuildHashOperator.hpp"
 #include "relational_operators/CreateIndexOperator.hpp"
@@ -282,6 +283,9 @@ void ExecutionGenerator::generatePlanInternal(
     case P::PhysicalType::kUpdateTable:
       return convertUpdateTable(
           std::static_pointer_cast<const P::UpdateTable>(physical_plan));
+    case P::PhysicalType::kWindowAggregate:
+      THROW_SQL_ERROR()
+          << "Window aggregate function is not supported yet :(";
     default:
       LOG(FATAL) << "Unknown physical plan node "
                  << physical_plan->getShortString();
--- a/query_optimizer/physical/WindowAggregate.hpp
+++ b/query_optimizer/physical/WindowAggregate.hpp
@@ -1,0 +1,133 @@
+/**
+ *   Copyright 2011-2015 Quickstep Technologies LLC.
+ *   Copyright 2015 Pivotal Software, Inc.
+ *   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
+ *     University of Wisconsin—Madison.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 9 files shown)


query_optimizer/ExecutionGenerator.cpp
```diff
@@ -1575,21 +1579,12 @@ void ExecutionGenerator::convertSort(const P::SortPtr &physical_sort) {
...
-  execution_plan_->addDirectDependency(
-      drop_merged_runs_index,
-      merge_run_operator_index,
-      true /* is_pipeline_breaker */);
-
+  // REVIEW(Shixuan): There might be an issue with the original code: if a throw
+  // happens before the plan is executed, merged_runs_relation will not be dropped
+  // by dropAllTemporaryRelations() if not added to temporary_relation_info_vec,
+  // which might cause "RelationNameCollision".
+  temporary_relation_info_vec_.emplace_back(merge_run_operator_index,
```
zuyu: @shixuan-fan I prefer the old way that deletes the tmp sorted relation eagerly, as the resultant blocks belonged to that relation would be huge.

No need to worry about the relation name collision, as it is created per query basis.

query_optimizer/physical/WindowAggregate.hpp
```diff
@@ -0,0 +1,133 @@
...
+
+  /**
+   * @brief Creates an WindowAggregate physical node.
+   *
+   * @param input The input node.
+   * @param window_aggregate_expression The window aggregate expression.
+   * @return An immutable WindowAggregate node.
+   */
+  static WindowAggregatePtr Create(
+      PhysicalPtr input,
```
zuyu: As we discussed [before](https://github.com/apache/incubator-quickstep/pull/39#discussion_r68329405), please use a `const` reference for `shared_ptr`.

query_optimizer/ExecutionGenerator.cpp
shixuan-fan: You are right. We should drop the table as early as possible. I previously thought the "DropTable" operator is inserted right after each temp_relation node. My bad.

However, When I tried to catch the physical plan for WindowAggregate using the old code, the first time it is okay, however, the second time it will report "RelationNameCollision". I believe the reason is that this temp relation is not dropped by dropAllTemporaryRelations() in the "try-catch". 

To handle the above issue, we need to add a check for WindowAggregate before visiting the children of each physical node, in order to avoid the creation of this temp relation (merged_runs_relation), as it will not be dropped after `THROW_SQL_ERROR()`.

zuyu: I think the actual executing order of `DropTable` depends on the scheduler based on a topological order. But in the optimizer generated DAG, a `sort` operator typically is placed as the last one, so in most case the new way is similar to the old way.

I suggest to debug the query w/ `window aggregation` by printing out the DAG and the operator finish order (hack in `query_execution/QueryManager`).

shixuan-fan: In the `window aggregation`, the window aggregation operator will be placed after the sort operator, that is, it will do the actual calculation after the table is sorted. So the work flow would be:

input_table -> sort -> window_aggregation -> output.

In my current implementation, there will be an exception at the physical node with `kWindowAggregate` type, which is preceded by a sort operator, and that's why `RelationNameCollision` happened.

zuyu: Then a quick fix is that add a true `pipeline_breaker` dependency from `window_aggregation` to `drop_merged_runs_index`.

shixuan-fan: Sounds good. Will revert to the original code. Thanks for all the discussion :)

PR #48, opened 2016-06-30
Handle negative number of worker threads as input

hbdeshmukh:
- Fixed a bug thereby Quickstep command line now handles negative
  number of worker threads correctly.
- If the user provides zero or fewer worker threads, we switch to the
  default number of worker threads, instead of terminating the process.


hbdeshmukh:
Assigning to @pateljm

rathijit:
This fixes the bug. Thanks @hbdeshmukh.

PR #52, opened 2016-07-07
Renamed Foreman related classes.

zuyu:
Renamed `ForemanBase` and `ForemanSingleNode` from `ForemanLite` and `Foreman`, respectively.

### DIFF
--- a/query_execution/CMakeLists.txt
+++ b/query_execution/CMakeLists.txt
@@ -1,6 +1,6 @@
 #   Copyright 2011-2015 Quickstep Technologies LLC.
 #   Copyright 2015-2016 Pivotal Software, Inc.
-#   Copyright 2016, Quickstep Research Group, Computer Sciences Department, 
+#   Copyright 2016, Quickstep Research Group, Computer Sciences Department,
 #     University of Wisconsin—Madison.
 #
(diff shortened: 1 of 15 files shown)


query_execution/CMakeLists.txt
```diff
@@ -82,12 +86,8 @@ target_link_libraries(quickstep_queryexecution_Foreman
...
                       quickstep_utility_EqualsAnyConstant
                       quickstep_utility_Macros
-                      tmb)
-target_link_libraries(quickstep_queryexecution_ForemanLite
-                      glog
-                      quickstep_threading_Thread
-                      quickstep_utility_Macros
-                      tmb)
+                      tmb
+                      ${GFLAGS_LIB_NAME})
```
hbdeshmukh: ``${GFLAGS_LIB_NAME}`` should appear before ``glog`` in alphabetical order.

hbdeshmukh:
LGTM except one minor style comment.

query_execution/CMakeLists.txt
zuyu: No, it starts w/ `$`, so it should be in the last.

See [here](https://github.com/apache/incubator-quickstep/blob/master/storage/CMakeLists.txt#L291) for example regarding `${PROTOBUF_LIBRARY}`.

PR #54, opened 2016-07-09
Introduced QueryManagerBase, and renamed QueryManagerSingleNode.

zuyu:


query_execution/QueryManagerBase.hpp
```diff
@@ -255,69 +212,75 @@ class QueryManager {
    *         otherwise.
    **/
   inline bool checkOperatorExecutionOver(const dag_node_index index) const {
-    if (checkRebuildRequired(index)) {
-      return (checkNormalExecutionOver(index) && checkRebuildOver(index));
-    } else {
-      return checkNormalExecutionOver(index);
-    }
+    return this->checkNormalExecutionOver(index) &&
+           (!checkRebuildRequired(index) || this->checkRebuildOver(index));
```
hbdeshmukh: I would argue that the original implementation is easier to read and debug.

zuyu: I would say that the refactor has the following logical flow: first check the normal execution of work orders, and then the rebuild.

query_execution/QueryManagerBase.hpp
```diff
@@ -255,69 +212,75 @@ class QueryManager {
...
-   * @brief Check if the rebuild operation for a given operator is over.
+   * @brief Fetch all work orders currently available in relational operator and
+   *        store them internally.
    *
-   * @param index The index of the given operator in the DAG.
+   * @param index The index of the relational operator to be processed in the
+   *        query plan DAG.
    *
-   * @return True if the rebuild operation is over, false otherwise.
+   * @return Whether any work order was generated by op.
```
hbdeshmukh: Better to say ``True if at least one work order was generate by the operator, false otherwise``.

query_execution/tests/QueryManager_unittest.cpp
```diff
@@ -28,7 +29,7 @@
 #include "query_execution/QueryExecutionMessages.pb.h"
 #include "query_execution/QueryExecutionState.hpp"
 #include "query_execution/QueryExecutionTypedefs.hpp"
-#include "query_execution/QueryManager.hpp"
+#include "query_execution/QueryManagerSingleNode.hpp"
```
hbdeshmukh: As we are testing only the QueryManagerSingleNode related functionality, we can rename the file to ``QueryManagerSingleNode_unittest.cpp``

hbdeshmukh:
HI @zuyu 

I have only few minor comments.

query_execution/QueryManagerBase.hpp
zuyu: Note that I did not change anything here, while `git` did the incorrect `diff` between `checkRebuildOver` and `fetchNormalWorkOrders`.

So, I prefer to leave it as is.

zuyu:
@hbdeshmukh I have renamed the `QueryManager` unit test as requested. Please take another look.

PR #56, opened 2016-07-13
Add a quick start guide.

pateljm:
Besides adding a quick start guide, also moved the old README to a developer README.

### DIFF
--- a/DEV_README.md
+++ b/DEV_README.md
@@ -1,0 +1,108 @@
+# Apache Quickstep (Incubating)
+
+[![Travis Widget]][Travis]
+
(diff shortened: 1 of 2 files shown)


zuyu:
LGTM, but next time when we try to rename a file, please use `git mv old-file new-file` so that `git diff` would present a better result, like [here](https://github.com/apache/incubator-quickstep/pull/54/files#diff-ff0a72a8f23829dedfbc638e047f9084).

zuyu:
@pateljm Never mind for the previous comment, but when squashing commits, we need to re-edit all commit messages.

PR #58, opened 2016-07-15
QUICKSTEP-20: Supported Window Aggregation for AVG

shixuan-fan:
Quickstep could now make the calculation for window aggregation function average. The work-flow is:

`ExecutionGenerator` constructs `WindowAggregationOperator` and initializes `WindowAggregationOperationState`
=> `WindowAggregationOperator::execute()` is called
=> `WindowAggregationOperationState::windowAggregate()`: combines all blocks' tuples into one `ValueAccessor`
=> `WindowAggregationHandle::calculate()`: actually does the calculation and return a `ColumnVector*` including all the calculated window aggregates
=> `WindowAggregationOperationState::windowAggregate()` will add the new `ColumnVector` into the created `ValueAccessor` and output the result to the `output_destination`.

We are using the simplest approach for the calculation: one thread will iterate through all the tuples. For each tuple, the thread will find all the elements in the given window and calculate the result.

In current version, we don't support `RANGE` frame mode. Also, other functions' handle will be created in the next PR.


expressions/window_aggregation/CMakeLists.txt
```diff
@@ -0,0 +1,212 @@
...
+add_library(quickstep_expressions_windowaggregation_WindowAggregateFunctionMax
+            WindowAggregateFunctionMax.cpp
+            WindowAggregateFunctionMax.hpp)
+add_library(quickstep_expressions_windowaggregation_WindowAggregateFunctionMin
+            WindowAggregateFunctionMin.cpp
+            WindowAggregateFunctionMin.hpp)
+add_library(quickstep_expressions_windowaggregation_WindowAggregateFunctionSum
+            WindowAggregateFunctionSum.cpp
+            WindowAggregateFunctionSum.hpp)
+add_library(quickstep_expressions_windowaggregation_WindowAggregateFunctionFactory
```
zuyu: Code style: please move it before `Max`.

shixuan-fan: Fixed! :)

expressions/window_aggregation/WindowAggregateFunction.hpp
```diff
@@ -0,0 +1,149 @@
...
+#include "expressions/window_aggregation/WindowAggregateFunction.pb.h"
+#include "expressions/window_aggregation/WindowAggregationID.hpp"
+#include "storage/StorageBlockInfo.hpp"
+#include "utility/Macros.hpp"
+
+namespace quickstep {
+
+class CatalogRelationSchema;
+class WindowAggregationHandle;
+class Type;
```
zuyu: Code style: Reorder declarations.

shixuan-fan: Done!

expressions/window_aggregation/CMakeLists.txt
zuyu: Unfortunately, I did not see the proposed changes.

expressions/window_aggregation/WindowAggregateFunctionAvg.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+bool WindowAggregateFunctionAvg::canApplyToTypes(
+    const std::vector<const Type*> &argument_types) const {
+  // AVG is unary.
+  if (argument_types.size() != 1) {
+    return false;
+  }
+
+  // Argument must be addable and divisible.
+  return BinaryOperationFactory::GetBinaryOperation(BinaryOperationID::kAdd)
```
zuyu: [Code style](https://google.github.io/styleguide/cppguide.html#Return_Values):

Please refactor to

```
return xxx &&
           yyy;
```

expressions/window_aggregation/WindowAggregateFunctionCount.cpp
```diff
@@ -0,0 +1,59 @@
...
+    const std::vector<const Type*> &argument_types) const {
+  if (!canApplyToTypes(argument_types)) {
+    return nullptr;
+  }
+
+  return &TypeFactory::GetType(kLong);
+}
+
+WindowAggregationHandle* WindowAggregateFunctionCount::createHandle(
+    std::vector<const Type*> &&argument_types,
```
zuyu: I am curious why we use the rvalue here?

expressions/window_aggregation/WindowAggregateFunctionCount.hpp
```diff
@@ -0,0 +1,75 @@
...
+#include <vector>
+
+#include "expressions/window_aggregation/WindowAggregateFunction.hpp"
+#include "expressions/window_aggregation/WindowAggregationID.hpp"
+#include "utility/Macros.hpp"
+
+namespace quickstep {
+
+class WindowAggregationHandle;
+class Type;
```
zuyu: Code style: Declaration ordering.

expressions/window_aggregation/WindowAggregateFunctionFactory.cpp
```diff
@@ -0,0 +1,106 @@
...
+  } else if (name == "sum") {
+    return &WindowAggregateFunctionSum::Instance();
+  } else {
+    return nullptr;
+  }
+}
+
+bool WindowAggregateFunctionFactory::ProtoIsValid(
+    const serialization::WindowAggregateFunction &proto) {
+  return proto.IsInitialized()
```
zuyu: [Code style](https://google.github.io/styleguide/cppguide.html#Return_Values):

Please refactor to
```
  return xxx &&
         yyy;
```

expressions/window_aggregation/WindowAggregateFunctionFactory.hpp
```diff
@@ -0,0 +1,102 @@
...
+ *        implementations of WindowAggregateFunction.
+ *
+ * WindowAggregateFunctionFactory allows client code to use any
+ * WindowAggregateFunction in Quickstep in a generic way without having to know
+ * about all the specific subclasses of WindowAggregateFunction. In particular,
+ * it is used to deserialize WindowAggregateFunctions used in
+ * WindowAggregationOperationState from their protobuf representations
+ * (originally created by the optimizer) when deserializing a QueryContext.
+ **/
+class WindowAggregateFunctionFactory {
```
zuyu: It is optional, and we could do the refactor later according to this [code style](https://google.github.io/styleguide/cppguide.html#Nonmember,_Static_Member,_and_Global_Functions).

We have to use the namespace install of static member functions.

expressions/window_aggregation/WindowAggregateFunctionMax.hpp
```diff
@@ -0,0 +1,75 @@
...
+#include <vector>
+
+#include "expressions/window_aggregation/WindowAggregateFunction.hpp"
+#include "expressions/window_aggregation/WindowAggregationID.hpp"
+#include "utility/Macros.hpp"
+
+namespace quickstep {
+
+class WindowAggregationHandle;
+class Type;
```
zuyu: Code style: reorder the declaration.

expressions/window_aggregation/WindowAggregateFunctionMin.hpp
```diff
@@ -0,0 +1,75 @@
...
+#include <vector>
+
+#include "expressions/window_aggregation/WindowAggregateFunction.hpp"
+#include "expressions/window_aggregation/WindowAggregationID.hpp"
+#include "utility/Macros.hpp"
+
+namespace quickstep {
+
+class WindowAggregationHandle;
+class Type;
```
zuyu: Code style: reorder the declaration.

expressions/window_aggregation/WindowAggregateFunctionSum.hpp
```diff
@@ -0,0 +1,75 @@
...
+#include <vector>
+
+#include "expressions/window_aggregation/WindowAggregateFunction.hpp"
+#include "expressions/window_aggregation/WindowAggregationID.hpp"
+#include "utility/Macros.hpp"
+
+namespace quickstep {
+
+class WindowAggregationHandle;
+class Type;
```
zuyu: Code style: reorder the declaration.

expressions/window_aggregation/WindowAggregationHandle.hpp
```diff
@@ -0,0 +1,137 @@
...
+                                  const std::int64_t num_following) const = 0;
+
+ protected:
+  /**
+   * @brief Constructor.
+   *
+   * @param partition_key_types The Types of the partition key.
+   **/
+  explicit WindowAggregationHandle(
+      std::vector<const Type*> &&partition_key_types) {
```
zuyu: I was wondering why using a rvalue here.

expressions/window_aggregation/tests/WindowAggregationHandleAvg_unittest.cpp
```diff
@@ -0,0 +1,398 @@
...
+        TypeFactory::GetType(typeID, static_cast<std::size_t>(10)) :
+        TypeFactory::GetType(typeID);
+
+    return WindowAggregateFunctionFactory::Get(WindowAggregationID::kAvg).canApplyToTypes(
+        std::vector<const Type*>(1, &type));
+  }
+
+  // Test resultTypeForArgumentTypes().
+  static bool ResultTypeForArgumentTypesTest(TypeID input_type_id,
+                                            TypeID output_type_id) {
```
zuyu: Code style: add one whitespace indentation.

expressions/window_aggregation/tests/WindowAggregationHandleAvg_unittest.cpp
```diff
@@ -0,0 +1,398 @@
...
+    const GenericType &type = GenericType::Instance(true);
+    initializeHandle(type);
+
+    // Create argument, partition key and cpptype vectors.
+    std::vector<typename GenericType::cpptype*> argument_cpp_vector;
+    argument_cpp_vector.reserve(kNumTuples);
+    ColumnVector *argument_type_vector =
+        createArgumentGeneric<GenericType>(&argument_cpp_vector);
+    NativeColumnVector *partition_key_vector =
+      new NativeColumnVector(IntType::InstanceNonNullable(), kNumTuples + 2);
```
zuyu: Code style: add two more whitespace indentation.

expressions/window_aggregation/tests/WindowAggregationHandleAvg_unittest.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+      typename GenericType::cpptype sum;
+      SetDataType(0, &sum);
+      sum += *argument_cpp_vector[i];
+      int count = 1;
+      bool is_null = false;
+
+      for (std::size_t precede = 1; precede <= kNumPreceding; ++precede) {
+        // Not the same partition.
+        if (i / kNumTuplesPerPartition != (i - precede) / kNumTuplesPerPartition
```
zuyu: [Code style](https://google.github.io/styleguide/cppguide.html#Boolean_Expressions):

Please refactor to
```
  if (xxx ||
      yyy)
```

expressions/window_aggregation/tests/WindowAggregationHandleAvg_unittest.cpp
```diff
@@ -0,0 +1,398 @@
...
+          break;
+        }
+
+        sum += *argument_cpp_vector[i - precede];
+        count++;
+      }
+
+      for (int follow = 1; follow <= kNumPreceding; ++follow) {
+        // Not the same partition.
+        if (i / kNumTuplesPerPartition != (i + follow) / kNumTuplesPerPartition
```
zuyu: Ditto for the code style for boolean expressions:

https://google.github.io/styleguide/cppguide.html#Boolean_Expressions

query_optimizer/expressions/WindowAggregateFunction.hpp
```diff
@@ -33,7 +33,7 @@
 
 namespace quickstep {
 
-class AggregateFunction;
+class WindowAggregateFunction;
```
zuyu: Code style: reorder the declaration.

query_optimizer/resolver/Resolver.cpp
```diff
@@ -2656,11 +2666,12 @@ E::ScalarPtr Resolver::resolveFunctionCall(
   }
 
   // Make sure a naked COUNT() with no arguments wasn't specified.
-  if ((aggregate->getAggregationID() == AggregationID::kCount)
-      && (resolved_arguments.empty())
-      && (!count_star)) {
-    THROW_SQL_ERROR_AT(&parse_function_call)
-        << "COUNT aggregate requires an argument (either scalar or star (*))";
+  if ((aggregate != nullptr && aggregate->getAggregationID() == AggregationID::kCount)
```
zuyu: Code style for the bool expressions:

https://google.github.io/styleguide/cppguide.html#Boolean_Expressions

query_optimizer/resolver/Resolver.cpp
```diff
@@ -2670,7 +2681,8 @@ E::ScalarPtr Resolver::resolveFunctionCall(
   }
 
   // Make sure that the aggregate can apply to the specified argument(s).
-  if (!aggregate->canApplyToTypes(argument_types)) {
+  if ((aggregate != nullptr && !aggregate->canApplyToTypes(argument_types))
```
zuyu: Code style for the bool expressions:

https://google.github.io/styleguide/cppguide.html#Boolean_Expressions

query_optimizer/tests/execution_generator/Select.test
```diff
@@ -953,19 +953,79 @@ WHERE double_col < 0
 ==
 
 # Window Aggregation Test.
-# Currently this is not supported, an empty table will be returned.
-SELECT avg(int_col) OVER w FROM test
+SELECT char_col, long_col, avg(long_col) OVER w FROM test
 WINDOW w AS
-(PARTITION BY char_col
- ORDER BY long_col DESC NULLS LAST
+(ORDER BY char_col DESC NULLS LAST
```
zuyu: Quick question: What if we remove the `order by` clause?

relational_operators/WindowAggregationOperator.cpp
```diff
@@ -67,6 +73,11 @@ serialization::WorkOrder* WindowAggregationOperator::createWorkOrderProto() {
   proto->set_query_id(query_id_);
   proto->SetExtension(serialization::WindowAggregationWorkOrder::window_aggr_state_index,
                       window_aggregation_state_index_);
+
+  std::vector<block_id> relation_blocks = input_relation_.getBlocksSnapshot();
```
zuyu: We could do some improvements on the code:

```
  const std::vector<block_id> relation_blocks = input_relation_.getBlocksSnapshot();
  for (const block_id bid : relation_blocks) { ... }
```

relational_operators/WindowAggregationOperator.hpp
```diff
@@ -58,16 +58,19 @@ class WindowAggregationOperator : public RelationalOperator {
...
    * @param query_id The ID of this query.
    * @param input_relation The relation to perform aggregation over.
+   * @param output_relation The relation for output.
    * @param window_aggregation_state_index The index of WindowAggregationState
    *                                       in QueryContext.
    * @param output_destination_index The index of InsertDestination in
    *                                 QueryContext for the output.
    **/
   WindowAggregationOperator(const std::size_t query_id,
+                            const CatalogRelation &input_relation,
```
zuyu: Remove this line?

relational_operators/WorkOrder.proto
```diff
@@ -249,6 +249,7 @@ message WindowAggregationWorkOrder {
   extend WorkOrder {
     // All required
     optional uint32 window_aggr_state_index = 336;
-    optional int32 insert_destination_index = 337;
+    repeated fixed64 block_ids = 337;
```
zuyu: FYI, we don't need to change the number assigned to `insert_destination_index`, but just add a new field for `block_ids` while assigning `338`.

storage/WindowAggregationOperationState.cpp
```diff
@@ -55,14 +62,12 @@ WindowAggregationOperationState::WindowAggregationOperationState(
     StorageManager *storage_manager)
     : input_relation_(input_relation),
       arguments_(std::move(arguments)),
-      partition_by_attributes_(std::move(partition_by_attributes)),
```
zuyu: Question: are we removing the support for the partitioned tuples?

storage/WindowAggregationOperationState.cpp
```diff
@@ -176,4 +171,122 @@ bool WindowAggregationOperationState::ProtoIsValid(const serialization::WindowAg
...
+  // work. This will need Order Key to be passed so that we know where the
+  // window should start and end.
+  if (!is_row_) {
+    std::cout << "Currently we don't support RANGE frame mode :(\n";
+    return;
+  }
+
+  // Get the total number of tuples.
+  int num_tuples = 0;
+  for (block_id block_idx : block_ids) {
```
zuyu: Add `const` for `block_id block_idx`.

storage/WindowAggregationOperationState.cpp
```diff
@@ -176,4 +171,122 @@ bool WindowAggregationOperationState::ProtoIsValid(const serialization::WindowAg
...
+    if (NativeColumnVector::UsableForType(type)) {
+      attribute_vecs.push_back(new NativeColumnVector(type, num_tuples));
+    } else {
+      attribute_vecs.push_back(new IndirectColumnVector(type, num_tuples));
+    }
+  }
+
+  // Construct column vectors for arguments.
+  std::vector<ColumnVector*> argument_vecs;
+  for (std::unique_ptr<const Scalar> &argument : arguments_) {
```
zuyu: Add `const` for `std::unique_ptr<const Scalar> &argument`.

storage/WindowAggregationOperationState.cpp
```diff
@@ -176,4 +171,122 @@ bool WindowAggregationOperationState::ProtoIsValid(const serialization::WindowAg
...
+    if (NativeColumnVector::UsableForType(type)) {
+      argument_vecs.push_back(new NativeColumnVector(type, num_tuples));
+    } else {
+      argument_vecs.push_back(new IndirectColumnVector(type, num_tuples));
+    }
+  }
+
+  // TODO(Shixuan): Add Support for Vector Copy Elision Selection.
+  // Add tuples and arguments into ColumnVectors.
+  for (block_id block_idx : block_ids) {
```
zuyu: Add `const` for `block_idx`.

storage/WindowAggregationOperationState.cpp
```diff
@@ -176,4 +171,122 @@ bool WindowAggregationOperationState::ProtoIsValid(const serialization::WindowAg
...
+    BlockReference block = storage_manager_->getBlock(block_idx, input_relation_);
+    const TupleStorageSubBlock &tuple_block = block->getTupleStorageSubBlock();
+    SubBlocksReference sub_block_ref(tuple_block,
+                                     block->getIndices(),
+                                     block->getIndicesConsistent());
+    ValueAccessor *tuple_accessor = tuple_block.createValueAccessor();
+    ColumnVectorsValueAccessor *argument_accessor = new ColumnVectorsValueAccessor();
+    for (std::unique_ptr<const Scalar> &argument : arguments_) {
```
zuyu: Add `const` for `argument`.

storage/WindowAggregationOperationState.proto
```diff
@@ -19,15 +19,15 @@ syntax = "proto2";
...
 message WindowAggregationOperationState {
-  required int32 relation_id = 1;
-  required AggregateFunction function = 2;
-  repeated Scalar arguments = 3;
-  repeated Scalar partition_by_attributes = 4;
-  required bool is_row = 5;
-  required int64 num_preceding = 6;  // -1 means UNBOUNDED PRECEDING.
-  required int64 num_following = 7;  // -1 means UNBOUNDED FOLLOWING.
+  required int32 input_relation_id = 1;
+  required WindowAggregateFunction function = 3;
```
zuyu: Something wrong w/ the number assignments and below.

storage/WindowAggregationOperationState.cpp
```diff
@@ -126,8 +121,8 @@ WindowAggregationOperationState* WindowAggregationOperationState::ReconstructFro
   const std::int64_t num_preceding = proto.num_preceding();
   const std::int64_t num_following = proto.num_following();
 
-  return new WindowAggregationOperationState(database.getRelationSchemaById(proto.relation_id()),
-                                             aggregate_function,
+  return new WindowAggregationOperationState(database.getRelationSchemaById(proto.input_relation_id()),
+                                             window_aggregate_function,
```
zuyu: FYI, we could directly use `&WindowAggregateFunctionFactory::ReconstructFromProto(proto.function())` for `window_aggregate_function`.

zuyu:
@shixuan-fan I've done one pass. Let me know if you want another.

expressions/window_aggregation/CMakeLists.txt
shixuan-fan: That's weird. Maybe I missed this commit during the squashing. Will fix it in the next commit.

expressions/window_aggregation/WindowAggregateFunctionCount.cpp
shixuan-fan: You are right. A reference would be better.

query_optimizer/tests/execution_generator/Select.test
shixuan-fan: Then the table will be in the same order as the input table when the window aggregates are calculated. And the calculated window aggregate for each row might depend on different values.

relational_operators/WindowAggregationOperator.cpp
shixuan-fan: Good catch!

storage/WindowAggregationOperationState.cpp
shixuan-fan: We will use attribute_id instead of Scalar.

storage/WindowAggregationOperationState.cpp
shixuan-fan: Sounds great.

shixuan-fan:
@zuyu Thanks for all your efforts! I modified the codes according to your suggestion. Your second pass will be most appreciated. Take your time, no hurry.

relational_operators/WorkOrder.proto
zuyu: So, do we want to refactor this changes?

query_optimizer/tests/execution_generator/Select.test
zuyu: I get it, but do we have other alternatives to keep the semantics of `window aggregation`: aggregating over incoming tuples.

It particular makes more sense for the following next test cases w/ `NULL`s.

relational_operators/WorkOrder.proto
shixuan-fan: Let's leave it as this to match the order in the fields of `WindowAggregationWorkOrder`.

query_optimizer/tests/execution_generator/Select.test
shixuan-fan: I don't quite understand. Could you please clarify this?

zuyu: I mean, we print the aggregation result according to the tuple order in the input table.

To that end, we may use a tuple id as the sort attribute.

shixuan-fan: Ah I get what you are saying. In that case we don't actually need to do a sort since nothing will be changed, and the sort attribute will be empty (though now we don't pass the vector into the handle).

The sort attribute is not needed currently since we only support ROWS mode. To support RANGE mode, we have to add sort attribute to calculate the window frame. They will be introduced in the future PR.

Thanks for pointing that out :)

zuyu: So, any proposed changes?

shixuan-fan: When the `RANGE` mode is introduced, a new parameter will be added to the `WindowAggregationHandle::calculate()` method, and would most likely be `const std::vector<attribute_id> &order_by_ids`. As for the cases where `ORDER BY` clause is missing, the vector would simply be empty, which indicates that the table is sorted by `tuple_id`. 

I don't think we need to explicitly add `tuple_id` as a sort attribute, because it is not actually an attribute.

shixuan-fan: @zuyu I found that the code logic here is a little bit wrong. For window w/ NULLs, the NULL values should be ignored rather than directly mark the result NULL. Only when all the values in the window are NULLs should the window aggregate result be marked NULL. 

Will fix this in the next commit.

shixuan-fan:
@zuyu I have a new version on my local repo which supports `RANGE`, and also optimizes the calculation from quadratic to linear. Could we have this one merged so that I could open another PR for the new feature? Thanks!

zuyu:
Wait for the CI.

shixuan-fan:
Sounds good. Thanks a lot! :)

PR #64, opened 2016-07-22
Introduced Shiftboss for the distributed version.

zuyu:


pateljm:
@zuyu Somewhere in the files, perhaps in QueryExecutionTypedefs.hpp, would be good to explain the RebuildWorkOrders highlighting when it is issued and the flow in which this is handled. Looks good otherwise. Thanks!

zuyu:
The `RebuildWorkOrder` info would be added in the next PR where `ForemanDistributed` is introduced.

query_execution/Shiftboss.hpp
```diff
@@ -0,0 +1,238 @@
...
+        start_worker_index_(0u) {
+    // Check to have at least one Worker.
+    DCHECK_GT(workers->getNumWorkers(), 0u);
+
+    shiftboss_client_id_ = bus_->Connect();
+    LOG(INFO) << "Shiftboss TMB client ID: " << shiftboss_client_id_;
+    DCHECK_NE(shiftboss_client_id_, tmb::kClientIdNone);
+
+    // Messages between Foreman and Shiftboss.
+    bus_->RegisterClientAsSender(shiftboss_client_id_, kShiftbossRegistrationMessage);
```
hbdeshmukh: I am unable to follow this - shouldn't we require two different types of buses? One for communication between foreman and shiftboss and another for communication between workers and shiftboss? I see only one bus instance being used here.

query_execution/Shiftboss.hpp
```diff
@@ -0,0 +1,238 @@
...
+    return shiftboss_client_id_;
+  }
+
+  /**
+   * @brief Get the Work Order processing capacity of all Workers managed by
+   *        Shiftboss during a single round of WorkOrder dispatch.
+   **/
+  inline std::size_t getWorkOrderCapacity() const {
+    DCHECK_NE(max_msgs_per_worker_, 0u);
+    return max_msgs_per_worker_ * workers_->getNumWorkers();
```
hbdeshmukh: If I understand this right, you are assuming a node wide limit for number of work orders which can be dispatched (let's call it N). There's another parameter which is the upper limit on the maximum on how many work orders can be dispatched to an individual worker (let's call it n). Assuming w workers, ``N = n * w``. Wouldn't it be better if we just configure a ``N`` and let the shiftboss decide how it needs to set ``n``? We may want to tackle it in the future, but adding a note here should be helpful to remind ourselves.

query_execution/Shiftboss.hpp
zuyu: That's a future work. For now, we use one TMB.

hbdeshmukh: Alright. Can you then rename the variables as ``internal_bus`` and ``external_bus`` appropriately for a better understanding? In the current implementation, they both will be equal to the same ``bus`` pointer.

query_execution/Shiftboss.hpp
zuyu: Since `w` is set, a configurable `n` has a fine-grain control over `N`.

query_execution/Shiftboss.hpp
zuyu: Not in this PR. Once we use two buses, the names make sense.

pateljm: I'd suggest we put a note here in this PR to outline what the design is. That can be augmented in the next PR. Thanks!

zuyu: @hbdeshmukh I've added [comments](https://github.com/apache/incubator-quickstep/pull/64/files#diff-d520fea60cec1aee1ead8dd4e6b91124R208) regarding two-level message buses for the communications.

query_execution/Shiftboss.cpp
```diff
@@ -0,0 +1,357 @@
...
+        CHECK(proto.ParseFromArray(tagged_message.message(), tagged_message.message_bytes()));
+
+        processInitiateRebuildMessage(proto.query_id(),
+                                      proto.operator_index(),
+                                      proto.insert_destination_index(),
+                                      proto.relation_id());
+        break;
+      }
+      case kWorkOrderCompleteMessage:  // Fall through.
+      case kRebuildWorkOrderCompleteMessage:
```
hbdeshmukh: Do you assume the following?
1. All the rebuild work orders are generated by the Foreman.
2. For every rebuild work order completion, a message is sent from Worker to Shiftboss, which is relayed to Foreman.

query_execution/Shiftboss.cpp
```diff
@@ -0,0 +1,357 @@
...
+
+        for (int i = 0; i < proto.blocks_size(); ++i) {
+          const block_id block = proto.blocks(i);
+          storage_manager_->saveBlockOrBlob(block);
+          if (storage_manager_->blockOrBlobIsLoaded(block)) {
+            // NOTE(zuyu): eviction is required to avoid accesses to the query
+            // result relation schema in CatalogDatabaseCache, for all query
+            // optimizer execution generator unit tests and the single-process
+            // Quickstep CLI.
+            storage_manager_->evictBlockOrBlob(block);
```
hbdeshmukh: I am not following this code and the corresponding NOTE - Why do you need to explicitly evict a block after calling ``saveBlockOrBlob`` on it?

query_execution/Shiftboss.cpp
```diff
@@ -0,0 +1,357 @@
...
+                  << "' message from worker (client " << annotated_message.sender << ") to Foreman";
+
+        DCHECK_NE(foreman_client_id_, tmb::kClientIdNone);
+        QueryExecutionUtil::SendTMBMessage(bus_,
+                                           shiftboss_client_id_,
+                                           foreman_client_id_,
+                                           move(annotated_message.tagged_message));
+        break;
+      }
+      case kQueryResultRelationMessage: {
```
hbdeshmukh: What is the purpose of this message? From the description in QueryExecutionTypedefs, it appears that the message gets sent from the Foreman to the Shiftboss. Why would a Shiftboss want to receive a result relation from the Foreman?

query_execution/Shiftboss.cpp
```diff
@@ -0,0 +1,357 @@
...
+
+size_t Shiftboss::getSchedulableWorker() {
+  const size_t num_workers = workers_->getNumWorkers();
+
+  size_t curr_worker = start_worker_index_;
+  for (;;) {
+    if (workers_->getNumQueuedWorkOrders(curr_worker) < max_msgs_per_worker_) {
+      start_worker_index_ = (curr_worker + 1) % num_workers;
+      // TODO(zuyu): workers_->incrementNumQueuedWorkOrders(curr_worker);
+      // But we need a WorkOrder queue first.
```
hbdeshmukh: It appears that you are doing round-robin style allocation of work to workers. If you do the increment as mentioned in your TODO comment above, you could pick a least loaded worker and assign it the work order. In that case you should also do a ``decrementNumQueuedWorkOrders()`` upon completion of a work order.

query_execution/Shiftboss.cpp
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    curr_worker = (curr_worker + 1) % num_workers;
+  }
+}
+
+void Shiftboss::registerWithForeman() {
+  LOG(INFO) << "Shiftboss (id '" << shiftboss_client_id_
+            << "') sent ShiftbossRegistrationMessage (typed '" << kShiftbossRegistrationMessage
+            << "') to all";
```
hbdeshmukh: Why should the registration message be broadcast? Are you assuming multiple Foremen?

query_execution/Shiftboss.cpp
```diff
@@ -0,0 +1,357 @@
...
+}
+
+void Shiftboss::processShiftbossInitiateMessage(
+    const std::size_t query_id,
+    const serialization::CatalogDatabase &catalog_database_cache_proto,
+    const serialization::QueryContext &query_context_proto) {
+  database_cache_.update(catalog_database_cache_proto);
+
+  unique_ptr<QueryContext> query_context(
+      new QueryContext(query_context_proto,
```
hbdeshmukh: By shift boss initiation, do you mean creation of a query context for a given query? Then the function and the message type names are misleading. Initiation of shiftboss could be interpreted as initialization of the shiftboss thread. Once the shiftboss is initialized, the Foreman can ask it to initialize the query context for a given query. I feel we should clarify this distinction.

query_execution/Shiftboss.cpp
zuyu: 1. No, but generated by each `Shiftboss`.
2. Yes.

query_execution/Shiftboss.cpp
```diff
@@ -0,0 +1,357 @@
...
+                                     shiftboss_client_id_,
+                                     foreman_client_id_,
+                                     move(ack_message));
+
+  for (size_t i = 0; i < partially_filled_block_refs.size(); ++i) {
+    // NOTE(zuyu): Worker releases the memory after the execution of
+    // RebuildWorkOrder on the Worker.
+    WorkOrder *rebuild_work_order =
+        new RebuildWorkOrder(query_id,
+                             move(partially_filled_block_refs[0]),
```
hbdeshmukh: Do you mean ``partially_filled_block_refs[i]``?

query_execution/Shiftboss.cpp
zuyu: `saveBlockOrBlob` does not evict a block from memory.

query_execution/Shiftboss.cpp
zuyu: Used to flush the query result.

query_execution/Shiftboss.hpp
```diff
@@ -0,0 +1,241 @@
...
+ *        the workorders to workers.
+ **/
+class Shiftboss : public Thread {
+ public:
+  /**
+   * @brief Constructor.
+   *
+   * @param bus A pointer to the TMB.
+   * @param storage_manager The StorageManager to use.
+   * @param workers A pointer to the WorkerDirectory.
```
hbdeshmukh: Who creates the ``WorkerDirectory`` which is being passed to the ShiftBoss? I believe that ShiftBoss itself should create it, as it controls the workers on that node.

query_execution/Shiftboss.cpp
zuyu: This function needs more refactor in many aspects, but I prefer to leave it as it to have a prototype of the distributed version running.

query_execution/Shiftboss.cpp
zuyu: No, but this is a way to allow `Shiftboss` to do a service discovery of `Foreman`.

zuyu: `Shiftboss` does not have a knowledge of where is `Foreman`.

query_execution/Shiftboss.cpp
hbdeshmukh: But why do you want the block to be evicted? What's the harm if it stays in memory, no one else modifies it AND you also have a copy on the disk (because of ``saveBlockOrBlob`` call)?

zuyu: See my comments above. It is because of the block is associated with a schema. And w/ eviction, it uses the schema from the catalog cache, which was invalid to the cli.

query_execution/Shiftboss.hpp
zuyu: No, there will be [an executable](https://github.com/apache/incubator-quickstep/blob/distributed-prototype/cli/WorkerDistributed.cpp#L190) to create `WorkerDirectory`, and `Shiftboss` is one thread of the process.

query_execution/Shiftboss.cpp
zuyu: Have changed to `QueryInitiate`. Thanks!

query_execution/Shiftboss.hpp
hbdeshmukh: Alright. Do you have a documentation describing these details? It's difficult to figure them out by looking at the code.

query_execution/QueryExecutionTypedefs.hpp
```diff
@@ -73,9 +73,17 @@ enum QueryExecutionMessageType : message_type_id {
...
+  kShiftbossRegistrationMessage,  // From Shiftboss to Foreman.
+  kShiftbossRegistrationResponseMessage,  // From Foreman to Shiftboss.
+  kQueryInitiateMessage,  // From Foreman to Shiftboss.
+  kQueryInitiateResponseMessage,  // From Shiftboss to Foreman.
+
   kInitiateRebuildMessage,  // From Foreman to Shiftboss.
   kInitiateRebuildResponseMessage,  // From Shiftboss to Foreman.
 
+  kQueryResultRelationMessage,  // From Foreman to Shiftboss.
+  kQueryResultRelationResponseMessage,  // From Shiftboss to Foreman.
```
hbdeshmukh: There are two kinds of message patterns: X to Shiftboss and Shiftboss to X. I can see that ShiftBoss constructor has appropriate type of registration related code. To appropriately register X with the bus, have you already written the code or you plan to address it in the next PR?

query_execution/Shiftboss.hpp
zuyu: `ForemanSingleNode` does not describe which creates `WorkerDirectory`. And it is obviously the `QuicksteCli` that does.

For the distributed case, the `Worker` executable does the similar task.

query_execution/QueryExecutionTypedefs.hpp
hbdeshmukh: Adding to the above comment: Because there are several types of messages, we can write a function which tests if all the sender and receivers are configured to send and receiver appropriate types of messages. That will give us an assurance that everything is correct in terms of communication.

zuyu: I think it would be addressed once `ForemanDistributed` is done.

Note that `Shiftboss` is working with the old query execution part w/o concurrent queries support. And now I am in the process of merging the working code w/ concurrent queries support.

query_execution/Shiftboss.cpp
hbdeshmukh: By looking at the code it seems that saveBlockOrBlob indeed writes a block to the disk. [Link](https://github.com/apache/incubator-quickstep/blob/master/storage/StorageManager.cpp#L406). So my previous comment still holds then.

query_execution/QueryExecutionTypedefs.hpp
zuyu: We will add a new unit test for the distributed query execution engine.

And I have tested it end-to-end in [a old repo](https://github.com/pivotalsoftware/quickstep/pull/117), w/o concurrent query support.

query_execution/Shiftboss.cpp
zuyu: What's your main comments again? There are many comments above.

The goal here is to remove the query result blocks from the buffer pool in the Worker site. which is remote to the distributed CLI.

query_execution/Shiftboss.cpp
hbdeshmukh: If the purpose of this message is to force the shiftboss to flush its query results, we should have a better name for the message type.

zuyu: Yes, I admitted that we would have a better name, and I'll fix it when adding `ForemanDistributed` in the next PR. Thanks!

query_execution/Shiftboss.cpp
```diff
@@ -0,0 +1,357 @@
...
+        break;
+      }
+      case kQueryResultRelationMessage: {
+        const TaggedMessage &tagged_message = annotated_message.tagged_message;
+
+        serialization::QueryResultRelationMessage proto;
+        CHECK(proto.ParseFromArray(tagged_message.message(), tagged_message.message_bytes()));
+
+        for (int i = 0; i < proto.blocks_size(); ++i) {
+          const block_id block = proto.blocks(i);
```
hbdeshmukh: The Foreman gives the ShiftBoss a list of block IDs belonging to the result relation, which the Shiftboss flushes to the disk, right? Two questions in this regard:

1. How does the Foreman get the list of block IDs?
2. Does the Foreman make sure that the block IDs sent to Shiftboss 1 has blocks belonging to Node 1 only?

query_execution/Shiftboss.cpp
hbdeshmukh: My point is that you need not evict the block explicitly. Mere ``saveBlockOrBlob`` call is enough.

query_execution/Shiftboss.cpp
zuyu: Note that the code works in the single `Worker` w/o concurrent query support, so things may change.

1. [`ForemanDistributed`](https://github.com/apache/incubator-quickstep/blob/distributed-exe-gen-test/query_execution/ForemanDistributed.cpp#L368) knows the query result relation from `QueryHandle`， which contains the blocks info.

2. For now, no. And we should refactor later.

hbdeshmukh: > Note that the code works in the single Worker

Do you mean single shiftboss? 

> ForemanDistributed knows the query result relation from QueryHandle， which contains the blocks info.

Does the ``ForemanDistributed`` get to know about blocks in the result relation via messages of type ``kCatalogRelationNewBlockMessage``? If so, who sends these messages to the ``ForemanDistributed``? I don't see this message type getting used in the ShiftBoss code. Then perhaps the workers? 

> For now, no. And we should refactor later.

This may cause issues when you extend the design to multiple shiftboss. Adding a TODO is highly recommended.

zuyu: Yes, single `Shiftboss`.

`kCatalogRelationNewBlockMessage` is used whenever `InsertDestination` creates a new block, and then notifies `Foreman` to add the new block into `Catalog`. And I believe `Shiftboss` would forward this message to `Foreman`.

`Foreman` knows the blocks from `QueryHandle`. Check out the link which contains the actual code.

This won't be a main issue, but just add some performance overhead for serialization.

query_execution/Shiftboss.cpp
zuyu: No, system will crash if we don't evict the blocks.

You could play with my old code in `distributed-exe-gen-test` to confirm my statement.

hbdeshmukh: > You could play with my old code in distributed-exe-gen-test to confirm my statement.

Please understand that I am not doubting what you have said. I am just trying to understand why do you need to evict a block from a result relation, which presumably won't be accessed by any other worker thread from that node at the time this line of code is getting executed. Your explanation from the source code reads as follows:

```
            // NOTE(zuyu): eviction is required to avoid accesses to the query
            // result relation schema in CatalogDatabaseCache, for all query
            // optimizer execution generator unit tests and the single-process
            // Quickstep CLI.
```

Why would optimizer execution generator unit tests interfere with the actual invocation of the system? 
Why does quickstep CLI cause an issue here? Is it something about concurrent accesses to a same block from two different processes?

query_execution/Shiftboss.cpp
hbdeshmukh: Are you planning to add a TODO in this PR then?

zuyu: No. Instead of adding a comment and waste few hours for the CI, I'll fix it in the next PR.

query_execution/Shiftboss.cpp
zuyu: Note that the new distributed query execution engine unit test is running in a process, w/ `Foreman`, `Shiftboss`, and `worker thread` running in different threads. It also shares `StorageManager`.

So evicting the query result blocks out from the buffer pool are required to avoid the reference to the schema in `CatalogDatabaseCache`. Using blocks in `cli` would result in invalid reference to the schema inaccessible to `cli`.

On the other hand, `cli` would load the blocks while using its own copy of `CatalogRelation`.

query_execution/Shiftboss.cpp
zuyu: FYI, there is no need for `TODO` in the `Shiftboss` side.

PR #65, opened 2016-07-25
Minor bug fix in QueryManager base.

hbdeshmukh:
Modified the order in which we check the completion of query and completion of an operator in the queryStatus function. Based on the semantics of the return value of ``queryStatus`` function, query completion return code takes precedence over operator completion. Hence changed the order in which we perform these checks. Note that this bug was present in the original code and was carried forward during the code refactoring of QueryManager.

### DIFF
--- a/query_execution/QueryManagerBase.cpp
+++ b/query_execution/QueryManagerBase.cpp
@@ -71,13 +71,14 @@ QueryManagerBase::QueryManagerBase(QueryHandle *query_handle)
 
 QueryManagerBase::QueryStatusCode QueryManagerBase::queryStatus(
     const dag_node_index op_index) {
-  if (query_exec_state_->hasExecutionFinished(op_index)) {
-    return QueryStatusCode::kOperatorExecuted;
-  }
-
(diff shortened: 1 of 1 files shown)


pateljm:
Looks good. @hbdeshmukh  Can you rebase?

hbdeshmukh:
HI @pateljm 

I have rebased with the master. Thanks.

zuyu:
The main question is that why our unit test does not catch this bug, and how to prevent similar bugs in the future (i.e, by adding a new unit test)?

hbdeshmukh:
The reason we did not encounter this bug was because there is only one place where this function is [called](https://github.com/apache/incubator-quickstep/blob/master/query_execution/PolicyEnforcerBase.cpp#L134). The call site is about removing a completed query and adding a waitlisted query. As there are no unit tests for multiple query execution yet, we did not catch this bug.

zuyu:
Actually, we should use `QueryExecutionState::hasQueryExecutionFinished` directly, and remove `queryStatus` method.

PR #63, opened 2016-07-22
DO NOT MERGE: Introduced PolicyEnforcer implementation for the distributed version.

zuyu:
We could hold until the distributed query execution unit test is done.

PR #69, opened 2016-07-28
DO NOT MERGE: Process InitRebuildResponse message.

zuyu:
For the distributed version.

### DIFF
--- a/query_execution/PolicyEnforcerBase.cpp
+++ b/query_execution/PolicyEnforcerBase.cpp
@@ -29,6 +29,7 @@
 #include "query_execution/QueryExecutionMessages.pb.h"
 #include "query_execution/QueryExecutionState.hpp"
 #include "query_execution/QueryManagerBase.hpp"
+#include "query_optimizer/QueryOptimizerConfig.h"  // For QUICKSTEP_DISTRIBUTED.
 #include "relational_operators/WorkOrder.hpp"
 #include "storage/StorageBlockInfo.hpp"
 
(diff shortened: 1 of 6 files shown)


query_execution/PolicyEnforcerBase.cpp
```diff
@@ -60,6 +61,23 @@ void PolicyEnforcerBase::processMessage(const TaggedMessage &tagged_message) {
       admitted_queries_[query_id]->processWorkOrderCompleteMessage(op_index);
       break;
     }
+#ifdef QUICKSTEP_DISTRIBUTED
```
hbdeshmukh: Why should we use this macro in the base class? Shouldn't this piece code go to the distributed derived class of the PolicyEnforcer?

query_execution/PolicyEnforcerBase.hpp
```diff
@@ -179,6 +180,18 @@ class PolicyEnforcerBase {
    **/
   virtual bool admitQuery(QueryHandle *query_handle) = 0;
 
+#ifdef QUICKSTEP_DISTRIBUTED
```
hbdeshmukh: Same comment as in PolicyEnforcerBase.cpp. Why are we using this macro in the base class? The base class can define virtual functions, as we wish. The single node derived implementation can be a no-op or through an error if the function concerns only with the distributed implementation.

query_execution/QueryManagerDistributed.cpp
```diff
@@ -119,6 +119,27 @@ bool QueryManagerDistributed::fetchNormalWorkOrders(const dag_node_index index)
...
+                                                                    const std::size_t num_rebuild_work_orders) {
+  // TODO(zuyu): Multiple workers support.
+  query_exec_state_->setRebuildStatus(op_index, num_rebuild_work_orders, true);
+
+  if (num_rebuild_work_orders != 0u) {
+    // Wait for the rebuild work orders to finish.
+    return;
+  }
+
+  markOperatorFinished(op_index);
```
hbdeshmukh: If we reach here, that means the operator did not issue any rebuild work orders, right? Adding that as an explanation should be helpful.

query_execution/QueryManagerDistributed.cpp
```diff
@@ -119,6 +119,27 @@ bool QueryManagerDistributed::fetchNormalWorkOrders(const dag_node_index index)
   return generated_new_workorder_protos;
 }
 
+void QueryManagerDistributed::processInitiateRebuildResponseMessage(const dag_node_index op_index,
+                                                                    const std::size_t num_rebuild_work_orders) {
+  // TODO(zuyu): Multiple workers support.
```
hbdeshmukh: Do you mean multiple shift bosses?

query_execution/PolicyEnforcerBase.hpp
zuyu: Because it also uses the code after `switch` for checking the status of a query execution.

query_execution/QueryManagerDistributed.cpp
zuyu: Yes, in the distributed version, without explicit explanations, `worker` means a worker process including `Shiftboss`, `StorageManager`, `CatalogDatabaseCache`, and a pool of `worker` threads.

query_execution/QueryManagerDistributed.cpp
zuyu: Yes, `num_rebuild_work_orders` is zero.

query_execution/QueryManagerDistributed.cpp
hbdeshmukh: Have we finalized this terminology somewhere? Using the same term ``worker`` in two contexts is confusing. I would recommend using an unambiguous term, like ``node``.

hbdeshmukh:
Hi @zuyu 

Couple of comments: 

1. I am not sure if I agree with the way in which inheritance is being applied here. The behavior of the base class should not be altered based on what the derived class is. 
2. Let's not use confusing terminologies.

zuyu:
@hbdeshmukh Thanks for the comments. Let us merge #63 first, and then revisit this PR. Thanks!

PR #73, opened 2016-08-01
Move hash join's probe and build node decision from ExecutionGenerator to PhysicalOptimizer

hakanmemisoglu:
Improvement for QUICKSTEP-37. ([https://issues.apache.org/jira/browse/QUICKSTEP-37]())

### DIFF
--- a/query_optimizer/physical/HashJoin.hpp
+++ b/query_optimizer/physical/HashJoin.hpp
@@ -20,6 +20,7 @@
 #ifndef QUICKSTEP_QUERY_OPTIMIZER_PHYSICAL_HASHJOIN_HPP_
 #define QUICKSTEP_QUERY_OPTIMIZER_PHYSICAL_HASHJOIN_HPP_
 
+#include <cstddef>
 #include <memory>
 #include <string>
 #include <type_traits>
--- a/query_optimizer/rules/BottomUpRule.hpp
+++ b/query_optimizer/rules/BottomUpRule.hpp
@@ -54,6 +54,7 @@ class BottomUpRule : public Rule<TreeType> {
   TreeNodePtr apply(const TreeNodePtr &tree) override {
     DCHECK(tree != nullptr);
 
+    init(tree);
     std::vector<std::shared_ptr<const TreeType>> new_children;
     bool has_changed_children = false;
     for (const std::shared_ptr<const TreeType> &child : tree->children()) {
@@ -79,6 +80,15 @@ class BottomUpRule : public Rule<TreeType> {
    * @param The updated node by the rule.
    */
   virtual TreeNodePtr applyToNode(const TreeNodePtr &node) = 0;
+
+  /**
+   * @brief Override this method to implement the initialization code
+   *        for the rule.
+   *
+   * @param input The input tree.
+   */
+  virtual void init(const TreeNodePtr &input) {
+  }
 
  private:
   DISALLOW_COPY_AND_ASSIGN(BottomUpRule);
--- a/query_optimizer/rules/SwapProbeBuild.cpp
+++ b/query_optimizer/rules/SwapProbeBuild.cpp
@@ -1,0 +1,62 @@
+#include "query_optimizer/rules/SwapProbeBuild.hpp"
+
+#include <cstddef>
+#include <memory>
+#include <vector>
+
+#include "query_optimizer/expressions/AttributeReference.hpp"
+#include "query_optimizer/physical/HashJoin.hpp"
+#include "query_optimizer/physical/PatternMatcher.hpp"
+#include "query_optimizer/physical/Physical.hpp"
+#include "query_optimizer/physical/TopLevelPlan.hpp"
+#include "query_optimizer/rules/Rule.hpp"
+
+
+namespace quickstep {
+namespace optimizer {
+
+P::PhysicalPtr SwapProbeBuild::applyToNode(const P::PhysicalPtr &input) {
+  P::HashJoinPtr hash_join;
+
+  if (P::SomeHashJoin::MatchesWithConditionalCast(input, &hash_join)
+      && hash_join->join_type() == P::HashJoin::JoinType::kInnerJoin) {
+    P::PhysicalPtr left = hash_join->left();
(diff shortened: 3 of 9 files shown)


query_optimizer/physical/HashJoin.hpp
```diff
@@ -172,18 +192,21 @@ class HashJoin : public BinaryJoin {
...
-        join_type_(join_type) {
+        join_type_(join_type),
+        estimated_right_cardinality_(estimated_right_cardinality) {
   }
 
   std::vector<expressions::AttributeReferencePtr> left_join_attributes_;
   std::vector<expressions::AttributeReferencePtr> right_join_attributes_;
   expressions::PredicatePtr residual_predicate_;
   JoinType join_type_;
+  std::size_t estimated_right_cardinality_;
```
jianqiao: It might not be good to couple the cardinality information here. We can still use `CostModel` to estimate a physical plan's cardinality whenever needed.

query_optimizer/rules/SwapProbeBuild.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+P::PhysicalPtr SwapProbeBuild::applyToNode(const P::PhysicalPtr &input) {
+  P::HashJoinPtr hash_join;
+
+  if (P::SomeHashJoin::MatchesWithConditionalCast(input, &hash_join)) {
+    P::PhysicalPtr left = hash_join->left();
+    P::PhysicalPtr right = hash_join->right();
+
+    P::TopLevelPlanPtr top_level;
+    if (P::SomeTopLevelPlan::MatchesWithConditionalCast(input, &top_level)) {
```
jianqiao: `input` should be already a `HashJoinPtr` here due to the outer `if` statement.

In fact, `BottomUpRule` does not provide a place for initializing `cost_model_`. One solution is to add an `initialization()` virtual method in `class BottomUpRule`, and implement that method in `SwapProbeBuild`.

query_optimizer/rules/SwapProbeBuild.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+  std::string getName() const override { return "SwapProbeBuild"; }
+
+ protected:
+  P::PhysicalPtr applyToNode(const P::PhysicalPtr &input) override;
+
+ private:
+  DISALLOW_COPY_AND_ASSIGN(SwapProbeBuild);
+
+  std::unique_ptr<C::StarSchemaSimpleCostModel> cost_model_;
```
jianqiao: Currently use `SimpleCostModel` instead of `StarSchemaSimpleCostModel` for deciding the build/probe sides.

zuyu: Actually, we should use the base class of the cost model, and initiate some implementation as @jianqiao suggested.

query_optimizer/physical/HashJoin.hpp
hakanmemisoglu: @jianqiao, I will remove it. Then I need to reintroduce the hashtable size estimation logic into ExecutionGenerator.

query_optimizer/rules/SwapProbeBuild.cpp
hakanmemisoglu: @jianqiao , Thanks for the catch. 

> In fact, BottomUpRule does not provide a place for initializing cost_model_.

It is the reason why, I have implemented initialization of cost model in "applyToNode". 
I will introduce init(const P::PhysicalPtr &top_level_plan) as you said.

query_optimizer/rules/BottomUpRule.hpp
```diff
@@ -80,6 +81,8 @@ class BottomUpRule : public Rule<TreeType> {
    */
   virtual TreeNodePtr applyToNode(const TreeNodePtr &node) = 0;
 
+  virtual void init(const TreeNodePtr &node) = 0;
```
jianqiao: Let's provide a default no-op implementation for `init()` here (instead of pure virtual) so we don't have to implement it for every subclass.

query_optimizer/rules/SwapProbeBuild.cpp
```diff
@@ -0,0 +1,64 @@
...
+P::PhysicalPtr SwapProbeBuild::applyToNode(const P::PhysicalPtr &input) {
+  P::HashJoinPtr hash_join;
+
+  if (P::SomeHashJoin::MatchesWithConditionalCast(input, &hash_join)
+      && hash_join->join_type() == P::HashJoin::JoinType::kInnerJoin) {
+    P::PhysicalPtr left = hash_join->left();
+    P::PhysicalPtr right = hash_join->right();
+
+
+
```
jianqiao: Remove the extra blank lines here (at most 1 line).

query_optimizer/rules/SwapProbeBuild.hpp
```diff
@@ -0,0 +1,47 @@
+#ifndef QUICKSTEP_QUERY_OPTIMIZER_RULES_SWAP_PROBE_BUILD_HPP_
+#define QUICKSTEP_QUERY_OPTIMIZER_RULES_SWAP_PROBE_BUILD_HPP_
+
+#include <string>
```
jianqiao: Add `#include <memory>` as `std::unique_ptr` is used in this file.

jianqiao:
LGTM except for some minor places. Also fix the style check issue then we're good to merge.

query_optimizer/rules/BottomUpRule.hpp
```diff
@@ -80,6 +81,14 @@ class BottomUpRule : public Rule<TreeType> {
    */
   virtual TreeNodePtr applyToNode(const TreeNodePtr &node) = 0;
 
+  /**
+   * @brief Initialization code to be used for each node.
```
jianqiao: The initialization is only applied with the tree's root node but not "each node".
Other places look good. I'll do this minor revision and then merge.

zuyu:
@jianqiao Per workflow, please delete the branch after merged. Thanks!

PR #74, opened 2016-08-02
Removed the redundant query id in the optimizer.

zuyu:


pateljm:
@jianqiao @hbdeshmukh Did any of you want the query_id for use in the future? 

@zuyu: Can you expand on the reason to drop the query_id further? 

Thanks!

zuyu:
`query_id` stores in `QueryHandle`, so `optimizer_context` does not need it.

And there will be some follow-up changes to `QueryHandle` to capture the `client_id` which submits the query.

jianqiao:
This change looks good to me.

PR #77, opened 2016-08-03
Added TMB Client Id in QueryHandle.

zuyu:
W/ the concurrent query support, `QueryHandle` needs to capture the TMB client id of the `CLI` which submits the query, so `Foreman` could later reply the query result to the right `CLI`.


pateljm:
LGTM. Merging.

PR #79, opened 2016-08-04
Constructed Generators once in the optimizer.

zuyu:
Instead of creating generators each time per query, do so once, and reuse the instances for every query.

### DIFF
--- a/query_optimizer/Optimizer.cpp
+++ b/query_optimizer/Optimizer.cpp
@@ -26,13 +26,11 @@ namespace optimizer {
 
 void Optimizer::generateQueryHandle(const ParseStatement &parse_statement,
                                     QueryHandle *query_handle) {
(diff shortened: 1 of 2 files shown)


zuyu:
@jianqiao I need your input on the correctness of this PR if we have concurrent queries running. Thanks!

jianqiao:
@zuyu The `Generator`'s can be stateless (need some refactoring), but each query should have its own `OptimizerContext`.

zuyu:
@jianqiao Thank you for your comments.

What do you think if we move `OptimizerContext` from `Optimizer` to `QueryHandle`, which is unique per query?

jianqiao:
That sounds good. Note that the refactoring needs quite some work to pass `optimizer_context` around, especially in `ExecutionGenerator`.

PR #81, opened 2016-08-04
Refactored QueryProcessor.

zuyu:
Assigned to @jianqiao.

This PR reverted `optimizer` per query, and avoids memory leaks if encountered any exceptions.


jianqiao:
LGTM. Merging.

PR #84, opened 2016-08-08
Added guards for gnu_source definition in storage CMakeLists

hbdeshmukh:
Background: We use mmap to allocate memory whenever it is available. There is a check in storage module's CMakeLists to further check if mmap can support HugePage allocation. 

Issue: The check for HugePage allocation used the definition for ``GNU_SOURCE`` macro. Some compilers by default have this macro defined, which causes the compiler to throw a warning "macro already defined". As warnings are treated as errors, this check fails, therefore CMake declares that HugePage support is not available. 

Fix: Guard the definition by #ifndef directive.


zuyu:
LGTM.

PR #88, opened 2016-08-09
Use glog for all sent TMB messages.

zuyu:


zuyu:
@hbdeshmukh Could you please take a look? Thanks!

hbdeshmukh:
Hi @zuyu 

The changes look fine. I have two comments though:

1. The amount of logging due to these changes will be a lot. If you need the logged information for debugging purposes, it may be worth to use something like DLOG or VLOG(3). 
2. If you need to do some analysis on the logs (e.g. counting number of messages sent from a particular client of a specified type), it may be worth to write a utility function which takes the following arguments:

- Sender TMB client ID
- Sender name string (e.g. ForemanSingle, ForemanDistributed, optional field)
- Reciver TMB client ID (some default value if this is a broadcast)
- Receiver name (optional)
- Message type
- Message type string
- 

The benefit of such a function is that you will get identical formatting for all the log messages. Therefore performing a grep over the logs will be much easier. Additionally, extending the function to log more information (e.g. message size, send/receive timestamps) will be straightforward.

zuyu:
@hbdeshmukh Thank you for your inputs. I'll use `DLOG` for this purpose.

As for the log analysis and the utility function, we could leave for future. For now, I just `glog` outputs for the debug purpose.

Cheers!

PR #85, opened 2016-08-08
Added ForemanDistributed and unit tests.

zuyu:

PR #93, opened 2016-08-14
Added ForemanDistributed.

zuyu:
Assigned to @hbdeshmukh. Thanks!

This is an initial version that accepts only one `Shiftboss`, tested by #94 while replaced #85.

### DIFF
--- a/query_execution/CMakeLists.txt
+++ b/query_execution/CMakeLists.txt
@@ -33,6 +33,9 @@ if (ENABLE_DISTRIBUTED)
   add_library(quickstep_queryexecution_BlockLocator BlockLocator.cpp BlockLocator.hpp)
 endif(ENABLE_DISTRIBUTED)
 add_library(quickstep_queryexecution_ForemanBase ../empty_src.cpp ForemanBase.hpp)
+if (ENABLE_DISTRIBUTED)
+  add_library(quickstep_queryexecution_ForemanDistributed ForemanDistributed.cpp ForemanDistributed.hpp)
+endif(ENABLE_DISTRIBUTED)
 add_library(quickstep_queryexecution_ForemanSingleNode ForemanSingleNode.cpp ForemanSingleNode.hpp)
 add_library(quickstep_queryexecution_PolicyEnforcerBase PolicyEnforcerBase.cpp PolicyEnforcerBase.hpp)
 if (ENABLE_DISTRIBUTED)
@@ -86,6 +89,26 @@ target_link_libraries(quickstep_queryexecution_ForemanBase
                       quickstep_threading_Thread
                       quickstep_utility_Macros
                       tmb)
+if (ENABLE_DISTRIBUTED)
+  target_link_libraries(quickstep_queryexecution_ForemanDistributed
+                        glog
+                        quickstep_catalog_CatalogDatabase
+                        quickstep_catalog_CatalogRelation
+                        quickstep_catalog_CatalogTypedefs
+                        quickstep_catalog_Catalog_proto
+                        quickstep_queryexecution_AdmitRequestMessage
+                        quickstep_queryexecution_ForemanBase
+                        quickstep_queryexecution_PolicyEnforcerDistributed
+                        quickstep_queryexecution_QueryExecutionMessages_proto
+                        quickstep_queryexecution_QueryExecutionTypedefs
+                        quickstep_queryexecution_QueryExecutionUtil
+                        quickstep_queryexecution_ShiftbossDirectory
+                        quickstep_threading_ThreadUtil
+                        quickstep_utility_EqualsAnyConstant
+                        quickstep_utility_Macros
+                        tmb
+                        ${GFLAGS_LIB_NAME})
+endif(ENABLE_DISTRIBUTED)
 target_link_libraries(quickstep_queryexecution_ForemanSingleNode
                       glog
                       quickstep_queryexecution_AdmitRequestMessage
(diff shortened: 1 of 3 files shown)


query_execution/CMakeLists.txt
```diff
@@ -33,6 +33,9 @@ if (ENABLE_DISTRIBUTED)
   add_library(quickstep_queryexecution_BlockLocator BlockLocator.cpp BlockLocator.hpp)
 endif(ENABLE_DISTRIBUTED)
 add_library(quickstep_queryexecution_ForemanBase ../empty_src.cpp ForemanBase.hpp)
+if (ENABLE_DISTRIBUTED)
+  add_library(quickstep_queryexecution_ForemanDistributed ForemanDistributed.cpp ForemanDistributed.hpp)
+endif(ENABLE_DISTRIBUTED)
```
hbdeshmukh: Can we have consistent behavior with the spacing between ``if (ENABLE_DISTRIBUTED)`` and ``endif(ENABLE_DISTRIBUTED)``?

zuyu: This is consistent in almost every, if not all, `CMakeLists.txt`s that `if (xxx)` has a whitespace , while all else, including `elseif(xxx)` and `else(xxx)`, do not have whitespace in between.

query_execution/ForemanDistributed.hpp
```diff
@@ -0,0 +1,130 @@
...
+namespace serialization { class WorkOrderMessage; }
+
+/** \addtogroup QueryExecution
+ *  @{
+ */
+
+/**
+ * @brief The Foreman receives queries from the main thread, messages from the
+ *        policy enforcer and dispatches the work to Shiftbosses. It also
+ *        receives work completion messages from Shiftbosses.
```
hbdeshmukh: Question: Does the brief description about Foreman class still hold true?

query_execution/ForemanDistributed.hpp
```diff
@@ -0,0 +1,130 @@
...
+   * @brief Send the given message to the specified worker.
+   *
+   * @param worker_index The logical index of the recipient worker in
+   *        ShiftbossDirectory.
+   * @param proto The WorkOrderMessage to be sent.
+   **/
+  void sendWorkOrderMessage(const std::size_t worker_index,
+                            const serialization::WorkOrderMessage &proto);
+
+  void processShiftbossRegisterationMessage(const tmb::client_id shiftboss_client_id,
```
hbdeshmukh: Grammar nitpick: Registration instead of Registeration.

query_execution/ForemanDistributed.cpp
```diff
@@ -0,0 +1,334 @@
...
+      case kCatalogRelationNewBlockMessage:  // Fall through
+      case kDataPipelineMessage:
+      case kRebuildWorkOrderCompleteMessage:
+      case kWorkOrderCompleteMessage:
+      case kWorkOrderFeedbackMessage: {
+        policy_enforcer_->processMessage(tagged_message);
+        break;
+      }
+      case kInitiateRebuildResponseMessage: {
+        // A unique case in the distributed version.
```
hbdeshmukh: What's the meaning of the comment: ``A unique case in the distributed version.``?

query_execution/ForemanDistributed.cpp
```diff
@@ -0,0 +1,334 @@
...
+        const MessageBus::SendStatus send_status =
+            bus_->Send(foreman_client_id_,
+                       shiftboss_addresses,
+                       broadcast_style,
+                       move(poison_message));
+        DCHECK(send_status == MessageBus::SendStatus::kOK);
+        return;
+      }
+      default:
+        LOG(FATAL) << "Unknown message type to Foreman";
```
hbdeshmukh: Here and at many other places in this file, can you rename ``Foreman`` to ``ForemanDistributed``? It's probably left over due to copying the function from the Foreman single node file.

query_execution/ForemanDistributed.cpp
```diff
@@ -0,0 +1,334 @@
...
+    }
+  }
+}
+
+bool ForemanDistributed::canCollectNewMessages(const tmb::message_type_id message_type) {
+  return !QUICKSTEP_EQUALS_ANY_CONSTANT(message_type,
+                                        kCatalogRelationNewBlockMessage,
+                                        kWorkOrderFeedbackMessage) &&
+         // TODO(zuyu): Multiple Shiftbosses support.
+         !shiftboss_directory_.hasReachedCapacity(0);
```
hbdeshmukh: Can we create a static  variable with name something like ``kShiftBossIndex`` to replace the ``0`` used here and elsewhere in the file?

query_execution/ForemanDistributed.hpp
zuyu: Yes.

query_execution/ForemanDistributed.cpp
zuyu: It means that this message is unique in the distributed version.

query_execution/ForemanDistributed.cpp
zuyu: No, alternatively once we have multiple shiftbosses support, it will be replaced with some variable called `shiftboss_index`.

query_execution/CMakeLists.txt
hbdeshmukh: Why should we rely on inconsistencies elsewhere in the project to introduce new ones? This is not a supremely important issue so I won't debate any further, however I don't agree with the reasons mentioned above.

zuyu:
I have addressed the comments. Please take another look. Thanks!

hbdeshmukh:
Merged.

zuyu:
Thanks!

PR #90, opened 2016-08-10
Quickstep 28 29

rathijit:
More efficient handling of aggregations by having a combined payload in the hash table. This also removes ValueT parametrization for the hash table data structure. Please refer to JIRA tickets QUICKSTEP-28 and QUICKSTEP-29 for more details.
TPC-H query 01 sees significant performance improvements with these changes.

Currently, only SeparateChaining hash table type is supported.

PR #96, opened 2016-08-23
QUICKSTEP-43 Introduced DestroyAggregationState operator

hbdeshmukh:
- Similar to the DestroyHash operator, this operator destroys the AggregationState once the Finalize aggregation operator finishes its execution.
- Optimizer support for DestroyAggregationState operator.

### DIFF
--- a/query_execution/QueryContext.hpp
+++ b/query_execution/QueryContext.hpp
@@ -182,16 +182,14 @@ class QueryContext {
   }
 
   /**
-   * @brief Release the given AggregationOperationState.
+   * @brief Destroy the given aggregation state.
    *
-   * @param id The id of the AggregationOperationState to destroy.
-   *
-   * @return The AggregationOperationState, alreadly created in the constructor.
+   * @param id The ID of the AggregationOperationState to destroy.
    **/
-  inline AggregationOperationState* releaseAggregationState(const aggregation_state_id id) {
+  inline void destroyAggregationState(const aggregation_state_id id) {
     DCHECK_LT(id, aggregation_states_.size());
     DCHECK(aggregation_states_[id]);
-    return aggregation_states_[id].release();
+    aggregation_states_[id].reset(nullptr);
   }
 
   /**
--- a/relational_operators/CMakeLists.txt
+++ b/relational_operators/CMakeLists.txt
@@ -33,6 +33,9 @@ add_library(quickstep_relationaloperators_AggregationOperator AggregationOperato
 add_library(quickstep_relationaloperators_BuildHashOperator BuildHashOperator.cpp BuildHashOperator.hpp)
 add_library(quickstep_relationaloperators_CreateIndexOperator CreateIndexOperator.cpp CreateIndexOperator.hpp)
 add_library(quickstep_relationaloperators_CreateTableOperator CreateTableOperator.cpp CreateTableOperator.hpp)
+add_library(quickstep_relationaloperators_DestroyAggregationStateOperator 
+            DestroyAggregationStateOperator.cpp 
+            DestroyAggregationStateOperator.hpp)
 add_library(quickstep_relationaloperators_DeleteOperator DeleteOperator.cpp DeleteOperator.hpp)
 add_library(quickstep_relationaloperators_DestroyHashOperator DestroyHashOperator.cpp DestroyHashOperator.hpp)
 add_library(quickstep_relationaloperators_DropTableOperator DropTableOperator.cpp DropTableOperator.hpp)
@@ -134,6 +137,16 @@ target_link_libraries(quickstep_relationaloperators_DeleteOperator
                       quickstep_storage_StorageBlockInfo
                       quickstep_storage_StorageManager
                       quickstep_threading_ThreadIDBasedMap
+                      quickstep_utility_Macros
+                      tmb)
+target_link_libraries(quickstep_relationaloperators_DestroyAggregationStateOperator
+                      glog
+                      quickstep_queryexecution_QueryContext
+                      quickstep_queryexecution_WorkOrderProtosContainer
+                      quickstep_queryexecution_WorkOrdersContainer
+                      quickstep_relationaloperators_RelationalOperator
+                      quickstep_relationaloperators_WorkOrder
+                      quickstep_relationaloperators_WorkOrder_proto
                       quickstep_utility_Macros
                       tmb)
 target_link_libraries(quickstep_relationaloperators_DestroyHashOperator
@@ -451,6 +464,7 @@ target_link_libraries(quickstep_relationaloperators_WorkOrderFactory
                       quickstep_relationaloperators_AggregationOperator
                       quickstep_relationaloperators_BuildHashOperator
                       quickstep_relationaloperators_DeleteOperator
+                      quickstep_relationaloperators_DestroyAggregationStateOperator
                       quickstep_relationaloperators_DestroyHashOperator
                       quickstep_relationaloperators_DropTableOperator
                       quickstep_relationaloperators_FinalizeAggregationOperator
@@ -484,6 +498,7 @@ target_link_libraries(quickstep_relationaloperators
                       quickstep_relationaloperators_CreateIndexOperator
                       quickstep_relationaloperators_CreateTableOperator
                       quickstep_relationaloperators_DeleteOperator
+                      quickstep_relationaloperators_DestroyAggregationStateOperator
                       quickstep_relationaloperators_DestroyHashOperator
                       quickstep_relationaloperators_DropTableOperator
                       quickstep_relationaloperators_FinalizeAggregationOperator
@@ -533,6 +548,7 @@ target_link_libraries(AggregationOperator_unittest
                       quickstep_queryexecution_QueryExecutionTypedefs
                       quickstep_queryexecution_WorkOrdersContainer
                       quickstep_relationaloperators_AggregationOperator
+                      quickstep_relationaloperators_DestroyAggregationStateOperator
                       quickstep_relationaloperators_FinalizeAggregationOperator
                       quickstep_relationaloperators_WorkOrder
                       quickstep_storage_AggregationOperationState_proto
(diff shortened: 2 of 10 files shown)


hbdeshmukh:
Some of the copyright blocks have been changed, I am fixing them now.

zuyu:
@hbdeshmukh Could you please give me the intuition of adding this new operator? Is it related to partition-aware aggregation? Thanks!

hbdeshmukh:
@zuyu I described the motivation for this operator in the JIRA issue, I will rephrase it here. 

>
Background: We use a class called AggregationState that manages the aggregation related data structures during the execution of a query. At present this class is destroyed through the FinalizeAggregation work order, which is executed by a single thread.
>
In the future, we would like to parallelize the FinalizeAggregation operator. Therefore for the destruction of the AggregationState, we need for a separate operator. This design pattern is similar to what is used in HashJoins, where we have a separate DestroyHashOperator.
>

zuyu:
@hbdeshmukh Could you please explain more regarding parallelizing the `FinalizeAggregation` operator?

zuyu:
Is it related to the `hash table pool`?

hbdeshmukh:
Sure. Right now we are using a single threaded implementation for the FinalizeAggregation operator. It does two things - 
1. Merging the multiple aggregation hash tables in one single global hash table. Note that same group by key can be present in different hash tables, hence the merging step is necessary.
2. Finalizing the aggregate value. e.g. in case of AVG, we delay the computation of actual average until very last. Until that point we keep the running sum and count.

There are multiple ways in which the FinalizeAggregation can be parallelized. One of them is divide the aggregation hash tables such that one hash table belongs to a given partition. We partition an input tuple using the group by key attributes and route that tuple to the appropriate hash table. Now, there are no overlapping keys across any two hash tables, hence the step 1 mentioned above is not necessary. As step 2 naturally relies on the output of 1, it can be done independently for each hash table. Therefore 1 and 2 can be parallelized.

query_execution/QueryContext.hpp
```diff
@@ -184,7 +184,7 @@ class QueryContext {
   /**
    * @brief Release the given AggregationOperationState.
    *
-   * @param id The id of the AggregationOperationState to destroy.
+   * @param id The id of the AggregationOperationState to release.
```
zuyu: I think we could remove this method, as it will not be used any more.

relational_operators/CMakeLists.txt
```diff
@@ -467,7 +481,6 @@ target_link_libraries(quickstep_relationaloperators_WorkOrderFactory
                       quickstep_relationaloperators_TableGeneratorOperator
                       quickstep_relationaloperators_TextScanOperator
                       quickstep_relationaloperators_UpdateOperator
-                      quickstep_relationaloperators_WindowAggregationOperator
```
zuyu: I was wondering why we removed this one?

relational_operators/DestroyAggregationStateOperator.hpp
```diff
@@ -0,0 +1,118 @@
...
+class DestroyAggregationStateOperator : public RelationalOperator {
+ public:
+  /**
+   * @brief Constructor.
+   *
+   * @param query_id The ID of the query to which this operator belongs.
+   * @param aggr_state_index The index of the AggregationState in QueryContext.
+   **/
+  DestroyAggregationStateOperator(const std::size_t query_id,
+                      const QueryContext::aggregation_state_id aggr_state_index)
```
zuyu: Code style: please align w/ the above line.

relational_operators/DestroyAggregationStateOperator.hpp
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * @brief Constructor.
+   *
+   * @param query_id The ID of the query to which this WorkOrder belongs.
+   * @param aggr_state_index The index of the AggregationState in QueryContext.
+   * @param query_context The QueryContext to use.
+   **/
+  DestroyAggregationStateWorkOrder(const std::size_t query_id,
+                       const QueryContext::aggregation_state_id aggr_state_index,
+                       QueryContext *query_context)
```
zuyu: Code style: Please align w/ `Line 96`.

relational_operators/WorkOrderFactory.cpp
```diff
@@ -420,22 +428,6 @@ WorkOrder* WorkOrderFactory::ReconstructFromProto(const serialization::WorkOrder
           shiftboss_client_id,
           bus);
     }
-    case serialization::WINDOW_AGGREGATION: {
```
zuyu: Again, why deleting `WindowAggr` code?

relational_operators/WorkOrderFactory.cpp
```diff
@@ -533,11 +530,13 @@ bool WorkOrderFactory::ProtoIsValid(const serialization::WorkOrder &proto,
         return false;
       }
 
+      const CatalogRelationSchema &build_relation = catalog_database.getRelationSchemaById(build_relation_id);
       const CatalogRelationSchema &probe_relation = catalog_database.getRelationSchemaById(probe_relation_id);
       for (int i = 0; i < proto.ExtensionSize(serialization::HashJoinWorkOrder::join_key_attributes); ++i) {
         const attribute_id attr_id =
             proto.GetExtension(serialization::HashJoinWorkOrder::join_key_attributes, i);
-        if (!probe_relation.hasAttributeWithId(attr_id)) {
+        if (!build_relation.hasAttributeWithId(attr_id) ||
```
zuyu: Note that I had removed the check for `build_relation`, as the `attr_id` in `probe_relation` may be different to that in `build_relation`, although both refer to the same `CatalogAttribute`.

relational_operators/WorkOrder.proto
```diff
@@ -42,6 +42,7 @@ enum WorkOrderType {
   TEXT_SCAN = 18;
   UPDATE = 19;
   WINDOW_AGGREGATION = 20;
+  DESTROY_AGGREGATION_STATE = 21;
```
zuyu: Please add a `TODO` to reorder the number based on the alphabet order.

relational_operators/WorkOrderFactory.cpp
hbdeshmukh: I forked this branch from the branch in #90, which in itself is quite old. Because of which some weird behavior may have happened. Thanks for pointing it out. I will remove any unrelated changes from this branch and let you know.

query_execution/QueryContext.hpp
hbdeshmukh: It's getting used in WorkOrderFactory.

hbdeshmukh:
Hi @zuyu 

Thanks for the comments. I have addressed them.

zuyu:
@hbdeshmukh If we build w/ the distributed version enabled using `-DENABLE_DISTRIBUTED=true`, we will observe multiple distributed-version-related unit tests failures due to sanity checks on [the Aggregation State](https://github.com/apache/incubator-quickstep/blob/destroy-agg-state-operator/query_execution/QueryContext.hpp#L204).

I would like to merge this PR after all the tests pass. I am looking into this issue, but you may also want to investigate on this. Thanks!

hbdeshmukh:
Hi @zuyu 

I am not sure what's causing the failures. The check is similar to other methods related to aggregation state. I am not quite familiar with the distributed version tests, so if you can let me know why this check is not appropriate, that could be helpful. 

On a side note, the QueryContext class lacks thread safety right now. For bigger query plans when several operators are under execution, there could be data races which can corrupt the class members.

zuyu:
From the initial observation, the reason of some query cases in the following unit tests failed seems that the aggregation state is already invalid before executing `DestroyAggregationStateWorkOrder`. I will take a further dig.

```
	 35 - quickstep_queryoptimizer_tests_distributed_executiongenerator_distinct (OTHER_FAULT)
	 37 - quickstep_queryoptimizer_tests_distributed_executiongenerator_index (OTHER_FAULT)
	 39 - quickstep_queryoptimizer_tests_distributed_executiongenerator_join (OTHER_FAULT)
	 40 - quickstep_queryoptimizer_tests_distributed_executiongenerator_select (OTHER_FAULT)
	 42 - quickstep_queryoptimizer_tests_distributed_executiongenerator_tablegenerator (OTHER_FAULT)
```

pateljm:
I'd suggest we close this PR and then look at the distributed version tests. Those were added in a hurry if I remember correctly (and without a PR). So how about @zuyu work on the distributed version fixes in a later PR.

query_execution/QueryContext.hpp
zuyu: @hbdeshmukh Since we changed the semantics of `FinalizeAggregationOperator`, when constructing `FinalizeAggregationWorkOrder`, [`WorkOrderFactory`](https://github.com/apache/incubator-quickstep/pull/96/files#diff-2c12385b34f03b5eb61a3f5df00505eeR157) should also use the get version instead. And it fixes the distributed unit tests. Otherwise, it would also cause memory leaks for the distributed version.

@pateljm The test failures are related to the missed changes in new feature for the distributed version, not my existing code. The main components in the distributed query execution engine was code reviewed, although not the test drivers (but the most driver code is forked from the existing codebase).

query_execution/QueryContext.hpp
```diff
@@ -195,6 +195,17 @@ class QueryContext {
   }
 
   /**
+   * @brief Destroy the given aggregation state.
+   *
+   * @param id The ID of the AggregationOperationState to destroy.
+   **/
+  inline void destroyAggregationState(const aggregation_state_id id) {
```
zuyu: As a side node for the new `DestroyAggregationStateWorkOrder`, we don't need this new method. Instead, using `releaseAggregationState` and the `unique_ptr` in the `WorkOrder` automatically destroys the content.

hbdeshmukh:
Hi @zuyu 

Applied your suggestions. Thanks.

> As a side node for the new DestroyAggregationStateWorkOrder, we don't need this new method. Instead, using releaseAggregationState and the unique_ptr in the WorkOrder automatically destroys the content.

This change caused some of the tests (Aggregation operator unit test and execution generator - select, distinct and join) to fail. I didn't follow up on why they were failing and stuck to using ``destroyAggregationState`` instead.

zuyu:
Hi @hbdeshmukh That's fine. Then, after we remove the unused `releaseAggregationState` in `QueryContext`, I would merge this PR.

hbdeshmukh:
Removed the ``releaseAggregationState`` method.

PR #97, opened 2016-08-31
QUICKSTEP-44 Separate Date type from Datetime type

hbdeshmukh:
- Beginning from the parser to the operator execution level, treat Date type
  sepearately than the Datetime type.
- Provide support for the extract function when applied on the Date
  type, implemented in the DateExtractOperation.
- Modified the tests.

Running the following query on master vs this branch on TPC-H SF100 dataset:

``SELECT EXTRACT(year FROM o_orderdate) AS orderyear FROM orders;``

| Master  | datetype-fix |
| ------------- | ------------- |
| 31700 ms  | 256 ms |

The reason for the performance improvement is because in the earlier implementation, a temporary date was created using the ``gmtime_r`` function call, which involves acquiring a mutex lock. In this branch, we don't need such locks as the extract function only works on the given date, stored in the storage blocks.


pateljm:
Sweet!

types/operations/unary_operations/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# Licensed to the Apache Software Foundation (ASF) under one
+# Licensed to the Agache Software Foundation (ASF) under one
```
zuyu: ?

pateljm:
With the fix to the spelling name above, shall we close this PR? @hbdeshmukh @zuyu Thanks!

zuyu:
Merged this PR, as it also passed the distributed unit tests.

PR #98, opened 2016-09-02
QUICKSTEP-53 DateLit improvements

hakanmemisoglu:
This PR changes DateLit represantation:

- Encode year, month and day information into an u32 field instead of keeping them separate in three different fields. The change reduces the size of DateLit from 48 bits to 32 bits.
- With the internal representation change, we changed the comparison operators' implementation. They now use one unsigned integer comparison for each method.

- Performance of the following query on TPCH data scale factor 100. 
Note: Before each query, `echo 3 > /proc/sys/vm/drop_caches && free && sync` command was run.
`SELECT EXTRACT(year FROM o_orderdate) AS orderyear FROM orders;`

|  Run    | master (ms)  | date-representation (ms) |
|----------|-----------------:|--------------------------------:|
|  1        |  34389.416    |   34850.132 
|  2        |  234.757        |   182.955 
|  3        |  221.087        |   177.373 
|  4        |  231.232        |   203.132
|  5        |  221.473        |  190.977

EDIT: Performance numbers are added.

### DIFF
--- a/types/DatetimeLit.hpp
+++ b/types/DatetimeLit.hpp
@@ -36,10 +36,13 @@ namespace quickstep {
  * @brief A literal representing the date.
  **/
 struct DateLit {
-  // Note that although there is no year 0 in the Gregorian calendar, ISO 8601
-  // has year 0 equivalent to 1 BCE, year -1 equivalent to 2 BCE, and so on.
-  std::int32_t year;
-  std::uint8_t month, day;
+  // ----------------------------------------------------
+  // |  23 bits                       | 4 bits | 5 bits |
+  // ----------------------------------------------------
+  // |       year                     |  month |   day  |
+  // | (unsigned)  (18 bits used)     |  (unsigned)     |
+  // ----------------------------------------------------
+  std::uint32_t year_month_day;
 
   // The maximum number of characters needed to represent any date in ISO 8601
   // notation.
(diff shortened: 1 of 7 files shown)


pateljm:
Nice observation and enhancement @hakanmemisoglu!

zuyu:
@hakanmemisoglu My 2 cents:

 1. IMHO, such `DateLit` encode is not necessary, as inside `TypedValue` it stores in a 64-byte `union` space.
 2. This PR shares many changes with #97. Thus, you may want to work w/ @hbdeshmukh to reduce the common changes.

hbdeshmukh:
Hi @zuyu 

>IMHO, such DateLit encode is not necessary, as inside TypedValue it stores in a 64-byte union space.
For the base table storage, your observation is correct. However, for evaluation of the EXTRACT and comparisons on the date, the compact representation should certainly be helpful.

zuyu:
@hbdeshmukh I guess the possible performance improvements come from a) using `DateLit` instead of `DatetimeLit`, b) a better implementation of `Extract` function, but not the internal representation of `DateLit`.

@hakanmemisoglu Could you please split the changes in this PR into multiple new PRs, and then compare the performance only between two different representation of `DateLit`? Thanks!

hakanmemisoglu:
@zuyu  This will be merged after @hbdeshmukh's changes. Since we are use rebasing, I don't think it will be a problem. 
For the efficiency part:
1-) Using new comparison operators obviously are faster than old counterparts using if else statements. (Of course the overall effect might not be fast as we expected in the TPCH queries.)  Thank you.
2-) No matter why TypedValue is used in query processing , the cutting the size of struct representation by 50% improves the memory usage:
 a-)  A page can contain 50% more values.
 b-) Recent x86 architectures favor using the values 32 bits or 64 bits, not 48 bits.

zuyu:
> For the efficiency part:
1-) Using new comparison operators obviously are faster than old counterparts using if else statements. (Of course the overall effect might not be fast as we expected in the TPCH queries.) Thank you.

I agree there would be some improvements in the theory, but do you have the number?

> 2-) No matter why TypedValue is used in query processing , the cutting the size of struct representation by 50% improves the memory usage:
a-) A page can contain 50% more values.
b-) Recent x86 architectures favor using the values 32 bits or 64 bits, not 48 bits.

I don't think so. The existing represent of one `uint32_t` and two `uint8_t`s would take 64 bits in total due to memory alignment requirements, instead of 48 bits as what you thought. Furthermore, as we use `TypeValue` instead of `DateLit` directly, I don't see much space improvements after refactoring using one single `uint32_t`.

hakanmemisoglu:
@zuyu I don't have the number. I'll post it when I have.

> I don't think so. The existing represent of one uint32_t and two uint8_ts would take 64 bits in total due to memory alignment requirements, instead of 48 bits as what you thought. 

Yes, I checked it, you are right. I did not know structs as themselves have a natural alignment. It will  be padded with 2 byte because first member.

However, this does not make my argument false, even supports it. As far as I know we store columns in natural format, then when we process them, we convert them into series of TypedValues. So keeping the sizes of structs small will help when we keep DateLit in storage memory.

zuyu:
@hakanmemisoglu Again, we rarely use `DateLit` directly. On the other hand, when using `TypedValue`, any changes to `DateLit` does not save memory space due to the union alignment inside `TypedValue`. In our case, it would always be 64 bits.

So I suggest we have the numbers before merging this PR.

types/operations/unary_operations/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# Licensed to the Apache Software Foundation (ASF) under one
+# Licensed to the Agache Software Foundation (ASF) under one
```
zuyu: ?

types/DatetimeLit.hpp
```diff
@@ -51,53 +54,70 @@ struct DateLit {
...
+
   static DateLit Create(const std::int32_t _year,
                         const std::uint8_t _month,
                         const std::uint8_t _day) {
     DateLit date;
-    date.year = _year;
-    date.month = _month;
-    date.day = _day;
+    // Normalize year by adding kMaxYear value, because we try to
+    // encode signed year value into an unsigned integer.
```
hbdeshmukh: Hi @hakanmemisoglu 

What's the impact on the leap year calculation because of this normalization (i.e. adding kMaxYear value to the negative value of year). More specifically, if ``_year`` is a leap year, is there a guarantee that ``_year + kMaxYear`` is also a leap year?

hakanmemisoglu: Hi @hbdeshmukh 
In the code, the leap year check and other checks such whether the year is valid or not, are done within `printValueToString() `and `parseValueFromString() `. These methods use `yearField()`, `monthField()`, `dayField()` to decode the real number from the representation. I think it should be OK. 
I did not understand actually why it should provide:
> if _year is a leap year, is there a guarantee that _year + kMaxYear is also a leap year? 

Or do I miss some other leap year checks done in the code?

hbdeshmukh: The check for leap year plays a role In the implementations for operator + and - for the date type. For instance, to perform an operation like ``-2000-29-01 + 13 months``, we need to know if the resultant year will be leap or not. It seems that this check for leap year will be performed on the normalized year instead of the original year.

hakanmemisoglu: **L125** `std::int32_t result_year = lhs.yearField() + (rhs.months / 12);`
**L126** `std::uint8_t result_month = static_cast<std::uint8_t>(lhs.monthField()) + (rhs.months % 12);`
**L134** `ClampDayOfMonth(result_year, result_month, lhs.dayField()));`

- Access to the struct happens via these methods (yearField(), monthField() etc...) that decodes from representation (normalized and shifted) to the year value (which is -2000 in your example). 
- After the operation + is done, and the check is applied. 
- The result year, month, day is calculated and given to `DateLit::Create()`
-  `DateLit::Create()`encodes the real values to the unified representation.

I might be missing something, but does the execution above make sense?

hbdeshmukh: Yes, it does. This means that the leap year calculations are indeed based on the original year and not the normalized one, which is correct behavior. Thanks for the clarification!

hakanmemisoglu:
I will close this PR since an issue in the compression code should be handled first.

PR #100, opened 2016-09-20
Refactor bulk insert for SplitRowStore

cramja:
This code refactors out multiple calls to the catalog in tight insert loops.

**Correctness**
I did not add any new tests because the tests in SplitRowStore_unittest covered both bulk and bulkWithRemappedAttributes insert methods.

**Benchmarks**
*Process*
Create a table with 6 int attributes and two varchar attributes. Then run a query with 50% selectivity and which selects on the varchar attribute as well as some of the int attributes, and insertion time should be slightly faster. Data in the relation is uniform-random distributed. The machine I tested this on was my laptop which should be a fair canidate because the relation in question is 1GB which easily fits in my memory.
```SQL
-- relation is ~ 1GB
create table splitrow (i0 int,i1 int,i2 int,i3 int,i4 int,s0 varchar(32),s1 varchar(32));
copy splitrow from 'splitrow.qs' WITH (DELIMITER '|');

-- test 1
select * from splitrow where i0 < 5000000; -- half the 1gb relation
select * from splitrow where i0 < 5000000;
select * from splitrow where i0 < 5000000;

-- test 2
select * from splitrow where i0 < 10; -- less than a one thousandth the 1gb relation
select * from splitrow where i0 < 10;
select * from splitrow where i0 < 10;
```

Test 1 results
```
Without change:   16420.618 ms
With change:       8122.209 ms
```

Test 2 results
```
Without change:   603.432 ms
With change:      604.629 ms
```
TLDR

We see a 2x improvement on large inserts.

### DIFF
--- a/storage/SplitRowStoreTupleStorageSubBlock.cpp
+++ b/storage/SplitRowStoreTupleStorageSubBlock.cpp
@@ -257,111 +257,13 @@ TupleStorageSubBlock::InsertResult SplitRowStoreTupleStorageSubBlock::insertTupl
 }
 
 tuple_id SplitRowStoreTupleStorageSubBlock::bulkInsertTuples(ValueAccessor *accessor) {
-  const tuple_id original_num_tuples = header_->num_tuples;
-  tuple_id pos = 0;
-
-  BasicInsertInfo insertInfo(relation_);
-
-  InvokeOnAnyValueAccessor(
-    accessor,
-    [&](auto *accessor) -> void {  // NOLINT(build/c++11
-      while (accessor->next()) {
-        // If packed, insert at the end of the slot array, otherwise find the
-        // first hole.
-        pos = this->isPacked() ? header_->num_tuples
-                               : occupancy_bitmap_->firstZero(pos);
-
-        // Only calculate space used if needed.
-        if (!this->spaceToInsert(pos, insertInfo.max_var_length_)) {
-          const std::size_t tuple_variable_bytes
-            = CalculateVariableSize<decltype(*accessor), true>(relation_, *accessor);
-          if (!this->spaceToInsert(pos, tuple_variable_bytes)) {
-            accessor->previous();
-            break;
-          }
-        }
-
-        // Find the slot and locate its sub-structures.
-        void *tuple_slot = static_cast<char *>(tuple_storage_) + pos * tuple_slot_bytes_;
-
-        BitVector<true> tuple_null_bitmap(tuple_slot, insertInfo.num_nullable_attrs_);
-        tuple_null_bitmap.clear();
-        char *fixed_length_attr_storage = static_cast<char *>(tuple_slot) + insertInfo.fixed_len_offset_;
-        std::uint32_t *variable_length_info_array =
-          reinterpret_cast<std::uint32_t *>(static_cast<char *>(tuple_slot) + insertInfo.var_len_offset_);
-
-        // Start writing variable-length data at the beginning of the
-        // newly allocated range.
-        std::size_t current_variable_position = tuple_storage_bytes_ - header_->variable_length_bytes_allocated;
-        std::uint32_t current_null_idx = 0;
-        for (attribute_id accessor_attr_id = 0; accessor_attr_id < insertInfo.num_attrs_; ++accessor_attr_id) {
-          bool nullable = insertInfo.is_nullable_.getBit(accessor_attr_id);
-          bool variable = insertInfo.is_variable_.getBit(accessor_attr_id);
-
-          if (!nullable && !variable) {
-            DCHECK_EQ(-1, relation_.getNullableAttributeIndex(accessor_attr_id));
-
-            const void *attr_value = accessor->template getUntypedValue<false>(accessor_attr_id);
-            std::memcpy(fixed_length_attr_storage + insertInfo.fixed_len_offsets_[accessor_attr_id],
-                        attr_value,
-                        insertInfo.fixed_len_sizes_[accessor_attr_id]);
-          } else if (nullable && !variable) {
-            DCHECK_EQ(relation_.getNullableAttributeIndex(accessor_attr_id), current_null_idx);
-
-            TypedValue attr_value(accessor->getTypedValue(accessor_attr_id));
-            if (attr_value.isNull()) {
-              tuple_null_bitmap.setBit(current_null_idx, true);
-            } else {
-              std::memcpy(fixed_length_attr_storage + insertInfo.fixed_len_offsets_[accessor_attr_id],
-                          attr_value.getDataPtr(),
-                          insertInfo.fixed_len_sizes_[accessor_attr_id]);
-            }
-            current_null_idx++;
-          } else if (!nullable && variable) {
-            TypedValue attr_value(accessor->getTypedValue(accessor_attr_id));
-
-            DCHECK_EQ(-1, relation_.getNullableAttributeIndex(accessor_attr_id));
-            DCHECK_EQ(insertInfo.var_len_offsets_[accessor_attr_id],
-                      relation_.getVariableLengthAttributeIndex(accessor_attr_id));
-            DCHECK(!attr_value.isNull());
-
-            const std::size_t attr_size = attr_value.getDataSize();
-            current_variable_position -= attr_size;
-            const int var_len_info_idx = insertInfo.var_len_offsets_[accessor_attr_id] * 2;
-            variable_length_info_array[var_len_info_idx] = current_variable_position;
-            variable_length_info_array[var_len_info_idx + 1] = attr_size;
-            attr_value.copyInto(static_cast<char *>(tuple_storage_) + current_variable_position);
-
-            header_->variable_length_bytes_allocated += attr_size;
-          } else {  // nullable, variable length
-            DCHECK_EQ(current_null_idx, relation_.getNullableAttributeIndex(accessor_attr_id));
-
-            TypedValue attr_value(accessor->getTypedValue(accessor_attr_id));
-            if (attr_value.isNull()) {
-              tuple_null_bitmap.setBit(current_null_idx, true);
-            } else {
-              DCHECK_EQ(relation_.getVariableLengthAttributeIndex(accessor_attr_id),
-                        insertInfo.var_len_offsets_[accessor_attr_id]);
-
-              const std::size_t attr_size = attr_value.getDataSize();
-              current_variable_position -= attr_size;
-              const int var_len_info_idx = insertInfo.var_len_offsets_[accessor_attr_id] * 2;
-              variable_length_info_array[var_len_info_idx] = current_variable_position;
-              variable_length_info_array[var_len_info_idx + 1] = attr_size;
-              attr_value.copyInto(static_cast<char *>(tuple_storage_) + current_variable_position);
-              header_->variable_length_bytes_allocated += attr_size;
-            }
-            current_null_idx++;
-          }
-        }
-        occupancy_bitmap_->setBit(pos, true);
-        ++(header_->num_tuples);
-        if (pos > header_->max_tid) {
-          header_->max_tid = pos;
-        }
-      }
-    });
-  return header_->num_tuples - original_num_tuples;
+  std::vector<attribute_id> simple_remap;
+  for (attribute_id attr_id = 0; 
+			attr_id < static_cast<attribute_id>(relation_.size());
+			++attr_id) {
+    simple_remap.push_back(attr_id);
+  }
+  return bulkInsertTuplesWithRemappedAttributes(simple_remap, accessor);
 }
 
 tuple_id SplitRowStoreTupleStorageSubBlock::bulkInsertTuplesWithRemappedAttributes(
@@ -385,7 +287,8 @@ tuple_id SplitRowStoreTupleStorageSubBlock::bulkInsertTuplesWithRemappedAttribut
         // Only calculate space used if needed.
         if (!this->spaceToInsert(pos, insertInfo.max_var_length_)) {
           const std::size_t tuple_variable_bytes
-            = CalculateVariableSizeWithRemappedAttributes<decltype(*accessor), true>(relation_, *accessor, attribute_map);
+            = CalculateVariableSizeWithRemappedAttributes<decltype(*accessor), true>(relation_, *accessor,
+                                                                                     attribute_map);
           if (!this->spaceToInsert(pos, tuple_variable_bytes)) {
             accessor->previous();
             break;
@@ -405,7 +308,8 @@ tuple_id SplitRowStoreTupleStorageSubBlock::bulkInsertTuplesWithRemappedAttribut
         // newly allocated range.
         std::size_t current_variable_position = tuple_storage_bytes_ - header_->variable_length_bytes_allocated;
         std::uint32_t current_null_idx = 0;
-        for (attribute_id accessor_attr_id = 0; accessor_attr_id < insertInfo.num_attrs_; ++accessor_attr_id) {
+        for (attribute_id accessor_attr_id = 0;
+             static_cast<std::size_t >(accessor_attr_id) < insertInfo.num_attrs_; ++accessor_attr_id) {
           bool nullable = insertInfo.is_nullable_.getBit(accessor_attr_id);
           bool variable = insertInfo.is_variable_.getBit(accessor_attr_id);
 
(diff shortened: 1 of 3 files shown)


cramja:
@navsan please take a look

storage/SplitRowStoreTupleStorageSubBlock.cpp
```diff
@@ -194,379 +257,125 @@ TupleStorageSubBlock::InsertResult SplitRowStoreTupleStorageSubBlock::insertTupl
...
+    accessor,
+    [&](auto *accessor) -> void {  // NOLINT(build/c++11
+      while (accessor->next()) {
+        // If packed, insert at the end of the slot array, otherwise find the
+        // first hole.
+        pos = this->isPacked() ? header_->num_tuples
+                               : occupancy_bitmap_->firstZero(pos);
+
+        // Only calculate space used if needed.
+        if (!this->spaceToInsert(pos, insertInfo.max_var_length_)) {
```
navsan: As discussed, let's move this function out of the loop by pre-calculating a conservative estimate of the available space. See my PackedRowStore bulkInsert implementation for example.

storage/SplitRowStoreTupleStorageSubBlock.cpp
```diff
@@ -194,379 +257,125 @@ TupleStorageSubBlock::InsertResult SplitRowStoreTupleStorageSubBlock::insertTupl
...
-      if (relation_.isVariableLength()) {
-        while (accessor->next()) {
-          pos = this->isPacked() ? header_->num_tuples
-                                 : occupancy_bitmap_->firstZero(pos);
+    accessor,
+    [&](auto *accessor) -> void {  // NOLINT(build/c++11
+      while (accessor->next()) {
+        // If packed, insert at the end of the slot array, otherwise find the
+        // first hole.
+        pos = this->isPacked() ? header_->num_tuples
```
navsan: This inner loop branch is why I didn't want to support the occupancy_bitmap_ at all. 

Any time there are "holes" in the tuple storage subblock, we would unset a number of bits in the occupancy_bitmap_. Then, during every bulk insert, we would go through a search through this large bitmap to get to the first zero. Is all this expense really worth it? Why not just disallow deletes (as PackedRowStore does), or allow holes to stay?

storage/SplitRowStoreTupleStorageSubBlock.cpp
```diff
@@ -194,379 +257,125 @@ TupleStorageSubBlock::InsertResult SplitRowStoreTupleStorageSubBlock::insertTupl
...
-                tuple_null_bitmap.setBit(nullable_idx, true);
-              } else {
-                std::memcpy(fixed_length_attr_storage
-                                + relation_.getFixedLengthAttributeOffset(attr_it->getID()),
-                            attr_value,
-                            attr_it->getType().maximumByteLength());
-              }
+
+        // Find the slot and locate its sub-structures.
+        void *tuple_slot = static_cast<char *>(tuple_storage_) + pos * tuple_slot_bytes_;
```
navsan: Can't we move this calculation out of the loop and keep just a tuple_slot increment here?

storage/SplitRowStoreTupleStorageSubBlock.cpp
```diff
@@ -194,379 +257,125 @@ TupleStorageSubBlock::InsertResult SplitRowStoreTupleStorageSubBlock::insertTupl
...
+
+        BitVector<true> tuple_null_bitmap(tuple_slot, insertInfo.num_nullable_attrs_);
```
navsan: SplitRowStore's null bitmap on a per-tuple basis is needlessly expensive. For instance, this constructor used here does a check on num_nullable_attrs to figure out how big the bitmap must be. That if/else statement certainly doesn't need to be done for every tuple insertion. 

In general, I think we should just move to having a single null bitmap for the entire subblock (i.e., across all tuples).

storage/SplitRowStoreTupleStorageSubBlock.cpp
```diff
@@ -88,6 +88,67 @@ inline std::size_t CalculateVariableSizeWithRemappedAttributes(
...
+  std::size_t num_attrs_;
+  std::size_t num_nullable_attrs_;
+  std::size_t max_var_length_;
+
+  // byte offset from the beginning of a tuple to the first fixed length attribute
+  std::uint32_t fixed_len_offset_;
+  // byte offset from the beginning of a tuple to the first variable length offset/length pair
+  std::uint32_t var_len_offset_;
+
+  BitVector<true> is_variable_;
```
navsan: Packing these Booleans into a bitvector is needlessly expensive. Just use straight up 1-byte bools instead. We only have a few columns, so the space overhead is negligible. But doing so will avoid the complexity/cost of bit arithmetic to do lookups during insertion.
[This applies to both the bitmaps.]

navsan: Actually, I think it's even better to just create one struct for each column, with all the info we care about for that column, like whether it's nullable/variable length and what its size is. That way, InsertInfo will have a vector<struct> which is better than a bunch of vector<int>s you have (because the struct packed with all the info you need can just be unpacked into registers directly, whereas the vector will force you to do memory loads for each column for each tuple insertion.

See my bulkInsert function in PackedRowStore.

navsan:
Please address the comments: mostly to do with performance improvements. Note that the objective here is not just to make SplitRowStore faster than it is today, but to get it to the point where we can drop PackedRowStore. We're not close to that right now, because we're doing far too much work in the inner loop. 

Also, 
- Consider getting rid of occupancy_bitmap_
- Consider moving the null_bitmap into the header as a single bitmap for the entire tuple storage subblock. 
- Consider getting rid of TypedValue

navsan:
Finally, can you also add support for partialBulkInsert functions? See the commits in my branch for reference. It'll help us improve the join result materialization cost, as well as a few other operations.

storage/SplitRowStoreTupleStorageSubBlock.cpp
```diff
@@ -194,379 +257,125 @@ TupleStorageSubBlock::InsertResult SplitRowStoreTupleStorageSubBlock::insertTupl
...
+            DCHECK_EQ(-1, relation_.getNullableAttributeIndex(accessor_attr_id));
+
+            const void *attr_value = accessor->template getUntypedValue<false>(attribute_map[accessor_attr_id]);
+            std::memcpy(fixed_length_attr_storage + insertInfo.fixed_len_offsets_[accessor_attr_id],
+                        attr_value,
+                        insertInfo.fixed_len_sizes_[accessor_attr_id]);
+          } else if (nullable && !variable) {
+            DCHECK_EQ(relation_.getNullableAttributeIndex(accessor_attr_id), static_cast<int>(current_null_idx));
+
+            TypedValue attr_value(accessor->getTypedValue(attribute_map[accessor_attr_id]));
```
navsan: You don't need to use a TypedValue here. Just check whether getUntypedValue returned a null pointer instead of using isNull.

storage/SplitRowStoreTupleStorageSubBlock.cpp
```diff
@@ -194,379 +257,125 @@ TupleStorageSubBlock::InsertResult SplitRowStoreTupleStorageSubBlock::insertTupl
...
+            if (attr_value.isNull()) {
```
navsan: It's safe to do this memcpy even if this value is null. Not important now, but useful when you want to coalesce memcpy-calls for contiguous attributes.

storage/SplitRowStoreTupleStorageSubBlock.cpp
```diff
@@ -194,379 +257,125 @@ TupleStorageSubBlock::InsertResult SplitRowStoreTupleStorageSubBlock::insertTupl
...
-            TypedValue attr_value(accessor->getTypedValue(*attr_map_it));
-            if (variable_idx != -1) {
-              const std::size_t attr_size = attr_value.getDataSize();
-              variable_length_info_array[variable_idx << 1] = current_variable_position;
-              variable_length_info_array[(variable_idx << 1) + 1] = attr_size;
-              attr_value.copyInto(static_cast<char*>(tuple_storage_) + current_variable_position);
-              current_variable_position += attr_size;
+            current_null_idx++;
+          } else if (!nullable && variable) {
+            TypedValue attr_value(accessor->getTypedValue(attribute_map[accessor_attr_id]));
```
navsan: Here, there's no way around using TypedValue right now. But please leave a note that we should change this to use a new (as yet unimplemented) getVariableLengthUntypedValue() function that returns both a void* pointer as well as a size.

storage/SplitRowStoreTupleStorageSubBlock.cpp
```diff
@@ -194,379 +257,125 @@ TupleStorageSubBlock::InsertResult SplitRowStoreTupleStorageSubBlock::insertTupl
...
+            DCHECK_EQ(insertInfo.var_len_offsets_[accessor_attr_id],
+                      relation_.getVariableLengthAttributeIndex(accessor_attr_id));
+            DCHECK(!attr_value.isNull());
+
+            const std::size_t attr_size = attr_value.getDataSize();
+            current_variable_position -= attr_size;
+            const int var_len_info_idx = insertInfo.var_len_offsets_[accessor_attr_id] * 2;
+            variable_length_info_array[var_len_info_idx] = current_variable_position;
+            variable_length_info_array[var_len_info_idx + 1] = attr_size;
+            attr_value.copyInto(static_cast<char *>(tuple_storage_) + current_variable_position);
```
navsan: This copyInto function uses bitshifts and has a needless branch in our codepath. Let's just use a memcpy with the pointer and attr_size known at this point.

storage/SplitRowStoreTupleStorageSubBlock.cpp
```diff
@@ -194,379 +257,125 @@ TupleStorageSubBlock::InsertResult SplitRowStoreTupleStorageSubBlock::insertTupl
...
+
+            header_->variable_length_bytes_allocated += attr_size;
+          } else {  // nullable, variable length
+            DCHECK_EQ(static_cast<int>(current_null_idx), relation_.getNullableAttributeIndex(accessor_attr_id));
```
navsan: Much of the code in this case is duplicated with the case above. Instead, just do the same thing as above (with the memcpy and no null-check) for all variable-length columns, and then do a null-check at the end for nullable ones. That way there's less code duplication, and almost exactly as much work (except in the pathological case of a large number of nulls).

storage/SplitRowStoreTupleStorageSubBlock.cpp
```diff
@@ -194,379 +257,125 @@ TupleStorageSubBlock::InsertResult SplitRowStoreTupleStorageSubBlock::insertTupl
...
+            DCHECK(!attr_value.isNull());
+
+            const std::size_t attr_size = attr_value.getDataSize();
+            current_variable_position -= attr_size;
+            const int var_len_info_idx = insertInfo.var_len_offsets_[accessor_attr_id] * 2;
+            variable_length_info_array[var_len_info_idx] = current_variable_position;
+            variable_length_info_array[var_len_info_idx + 1] = attr_size;
+            attr_value.copyInto(static_cast<char *>(tuple_storage_) + current_variable_position);
+
+            header_->variable_length_bytes_allocated += attr_size;
```
navsan: I think it's cleaner (and slightly more efficient) to update the header at the end, and just using current_variable_position as the running position index in the loop.

storage/SplitRowStoreTupleStorageSubBlock.cpp
```diff
@@ -194,379 +257,125 @@ TupleStorageSubBlock::InsertResult SplitRowStoreTupleStorageSubBlock::insertTupl
...
-                            + relation_.getFixedLengthAttributeOffset(attr_it->getID()),
-                        attr_value,
-                        attr_it->getType().maximumByteLength());
-          }
-          occupancy_bitmap_->setBit(pos, true);
-          ++(header_->num_tuples);
-          if (pos > header_->max_tid) {
-            header_->max_tid = pos;
-          }
+        occupancy_bitmap_->setBit(pos, true);
```
navsan: All of this would go away if we drop the occupancy_bitmap_. 

On the other hand, if we absolutely want to allow deletes, it'd be faster (and more space-efficient) to keep a sorted list of tuple IDs rather than an entire bitmap.

storage/SplitRowStoreTupleStorageSubBlock.cpp
cramja: > That if/else statement certainly doesn't need to be done for every tuple insertion

Which if/else? In my eyes this is not a branch at all.

> In general, I think we should just move to having a single null bitmap for the entire subblock (i.e., across all tuples).

What's the logic here. I think for predicate evaluation it makes sense to keep the null data 'close' to the attribute data, as they will likely be accessed at nearly the same time (ie in the creation of a TypedValue for predicate evaluation). For copying, it might be faster, but I'm not sure why necessarily, maybe you could explain.

storage/SplitRowStoreTupleStorageSubBlock.cpp
cramja: That's right. I will correct this.

(Should a compiler be able to infer this?)

storage/SplitRowStoreTupleStorageSubBlock.cpp
cramja: I like it, we can do that.

storage/SplitRowStoreTupleStorageSubBlock.cpp
cramja: Hmm, I agree this is an overly checky-check. However, I'm not going to address removing the delete functionality in this PR. That change would be large and not directly related to this change and so can wait for another PR. 

What I can do is to remove the check and just continuously increment `pos` like you suggest. Filling in gaps can still be accomplished by using a non-bulk `insertTuple` or calling rebuild so no functionality will be lost.

storage/SplitRowStoreTupleStorageSubBlock.cpp
cramja: Good catch. Initially I didn't do this because I thought it wouldn't be handled by the Compressed/ValueAccessors. However, scanning their code it seems like this will work, and I will make this change.

storage/SplitRowStoreTupleStorageSubBlock.cpp
cramja: Yes but we'll still need to set the null bit anyways so I don't see a change here.

storage/SplitRowStoreTupleStorageSubBlock.cpp
cramja: Okay, I will add `TODOs`

storage/SplitRowStoreTupleStorageSubBlock.cpp
cramja: Okay, sounds good.

storage/SplitRowStoreTupleStorageSubBlock.cpp
cramja: One less addition! Gotta catch em all™ :)

Part of the reason they do it incrementally is the dependence of method `spaceToInsert` on the header's information. 

I will see if this change makes sense after refactoring the related code, and if not leave a comment.

storage/SplitRowStoreTupleStorageSubBlock.cpp
cramja: Agreed that it would be smaller in the common case (empty!) and since the 
```c++
  std::unique_ptr<BitVector<false>> occupancy_bitmap_;
```
is on the heap anyways, it's not like we need to worry about how to serialize a variable length list when a block is evicted.

cramja:
@navsan love the comments, very thorough.

I updated the header of this PR to include benchmarks/testing info and will keep this in mind in the future.

**Changes queued for this PR**
* getting rid of TypedValue
* moving some of the computations out of the inner loop and keeping a leaner set of running state variables
* moving to a per-column info struct rather than a number of bitvectors and vector<int>s

I'm going to address these in an addendum commit. I think your suggestions are great.

**Different PR**
* getting rid of occupancy_bitmap_
* moving the null_bitmap into the header as a single bitmap for the entire tuple storage subblock.

I think these are valid issues (comments above for specifics) but that they can be the subject of another commit/PR in the pursuit of more concise PRs.

**Another PR**
> Can you coalesce writes for contiguous columns in input/output? This helps speed up the common case of materialized join results where (almost) everything gets copied out anyway.

> Can you also add support for partialBulkInsert functions? See the commits in my branch for reference. It'll help us improve the join result materialization cost, as well as a few other operations.

Both these points can be addressed but in a separate PR.

navsan:
Sounds good! Thanks.

cramja:
@navsan I made the following changes:
* Make per-column insert info
* Moved some variables outside the loop
* sped up the way that the block estimates the number of tuples it can insert

After these changes, the experiment (mentioned in the header to this PR) speeds up to `7267.457 ms` with the same setup.

pateljm:
@cramja This is a nice feature. Thanks @navsan for the detailed comments. 

Both, can we close this PR out over the next few days. Gotta keep moving ™ :)

cramja:
Agreed. I've been working on adding @navsan 's idea of merging contiguous attributes. I have another [branch with those changes](https://github.com/cramja/incubator-quickstep/tree/splitrow_partial_ins). It's almost ready, I'm in the process of benchmarking with TPCH to see if it can replace PackedRow.

Granted this PR speeds up Splitrow, and has tests so can be merged.

cramja:
I'm closing this PR in favor of another which uses a technique to merge attribute copies when possible.

PR #101, opened 2016-09-21
Optimize PackedRowStoreValueAccessor & BasicColumnStoreValueAccessor by removing redundant computations and clearly exposing a strided memory access pattern

saketj:
This PR proposes to optimize the way value accessors are currently used to read values from a storage block. Code profiling revealed that almost 10%-15% of the execution time was being spent in making redundant method calls to CatalogRelationSchema to find out properties that are anyway always constant when iterating over a given column. A simple caching of these values could give significant improvements. Moreover, during code analysis it was found that even more opportunities for compiler induced optimization and vectorization exist, if a simple semantics of a strided memory access pattern is exposed to the compiler & the runtime. After all, most iterations over a column in a relation can be expressed by as simple equation as: `base_address + tuple_id * offset`, where base_address and offset are constant. Exposing this semantics opens a whole lot of interesting opportunities, which otherwise have been obscured in the existing design of ValueAccessors. Rewriting the entire set of ValueAccessors is a mammoth  task. Therefore, this PR and few subsequent PRs will propose to do this incrementally, always making sure that each PR addresses end-to-end functionality for some subset of features.

This PR specifically optimizes PackedRowStoreValueAccessor & BasicColumnStoreValueAccessor only. To demonstrate the end-to-end benefit of this optimization, predicate evaluation involving relational operators, regular expressions & literal comparison have been refactored to be cognizant of such optimizations. 

Tests:
------
- Tests for changes to PackedRowStoreValueAccessor have been added in PackedRowStoreTupleStorargeSubBlock_unittest.cpp
- BasicColumnStoreTupleStorageSubBlock_unittest.cpp has existing tests which test various code paths that invoke the new functionalities for BasicColumnStoreValueAccessor.

Performance Numbers:
-----------------------

A simple selection query `select count(*) from lineorder where lo_discount > 5` on SSB Scale Factor 100 improves by almost 18-20%. Following lists the **query completion time** for 20 trials for SSB-SF-100 done on the standard Cloudlab machine with and without the current patch:

| Trial #       | w/o ColumnAccessors  (in seconds)         | w/ ColumnAccessor (in seconds) |
| ------------- |:-------------:| -----:|
| 1 | 293 | 218
| 2 | 267 | 222
| 3 | 254 | 227
| 4 | 255 | 230
| 5 | 256 | 193
| 6 | 244 | 214
| 7 | 244 | 207
| 8 | 235 | 198
| 9 | 226 | 189
| 10 | 225 | 182
| 11 | 226 | 184
| 12 | 223 | 181
| 13 | 222 | 177
| 14 | 224 | 174
| 15 | 225 | 173
| 16 | 219 | 170
| 17 | 219 | 161
| 18 | 220 | 164
| 19 | 217 | 161
| 20 | 219 | 159
| Mean |236 |189 |

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,5 @@
 Makefile.in
 autom4te.cache
 .DS_Store
+.idea
 *~
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ env:
   - BUILD_TYPE=Release VECTOR_COPY_ELISION_LEVEL=selection
 
 install:
-  - if [ "$CC" = "gcc" ]; then
+  - if [ "$CC" = "gcc" ] || [[ "$BUILD_TYPE" = "Release" &&  "$VECTOR_COPY_ELISION_LEVEL" = "selection" ]]; then
       export MAKE_JOBS=1;
     else
       export MAKE_JOBS=2;
--- a/expressions/scalar/ScalarAttribute.cpp
+++ b/expressions/scalar/ScalarAttribute.cpp
@@ -100,18 +100,43 @@ ColumnVector* ScalarAttribute::getAllValues(ValueAccessor *accessor,
                                                           accessor->getNumTuples());
       accessor->beginIteration();
       if (result_type.isNullable()) {
-        while (accessor->next()) {
-          const void *value = accessor->template getUntypedValue<true>(attr_id);
-          if (value == nullptr) {
-            result->appendNullValue();
-          } else {
-            result->appendUntypedValue(value);
+        if (accessor->isColumnAccessorSupported()) {
+          // If ColumnAccessor is supported on the underlying accessor, we have a fast strided
+          // column accessor available for the iteration on the underlying block.
+          // Since the attributes can be null, ColumnAccessor template takes a 'true' argument.
+          std::unique_ptr<const ColumnAccessor<true>>
+              column_accessor(accessor->template getColumnAccessor<true>(attr_id));
+          while (accessor->next()) {
+            const void *value = column_accessor->getUntypedValue();  // Fast strided access.
+            if (value == nullptr) {
+              result->appendNullValue();
+            } else {
+              result->appendUntypedValue(value);
+            }
+          }
+        } else {
+          while (accessor->next()) {
+            const void *value = accessor->template getUntypedValue<true>(attr_id);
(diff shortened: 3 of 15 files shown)


saketj:
@navsan Could you please review?

pateljm:
@saketj Can you restart Travis to make sure the tests come back green. 
@navsan If this looks good to you, can we close this? 
Thanks!

saketj:
@pateljm Thanks Jignesh. Have updated the patch and also added the performance numbers in the PR description.

saketj:
It is strange that only one of the Travis build tasks for [clang++ compiler, build=release and vector_copy_ellision_level=selection] is repeatedly failing due to a compiler error.  (Reproduced below) . 

The same code compiles perfectly for clang++ release (vector_copy_ellision_level=joinwithbinaryexpression) and for all configurations of g++. 

I don't think it is related to changes of current PR. I tried running it multiple times, but still the same error. Does anyone else has also faced the same issue before, or might know what is causing it?

> Building CXX object types/operations/comparisons/CMakeFiles/quickstep_types_operations_comparisons_GreaterComparison.dir/GreaterComparison.cpp.o
error: unable to execute command: Killed
clang: error: clang frontend command failed due to signal (use -v to see invocation)
Ubuntu clang version 3.7.1-svn253571-1~exp1 (branches/release_37) (based on LLVM 3.7.1)
Target: x86_64-pc-linux-gnu
Thread model: posix
clang: note: diagnostic msg: PLEASE submit a bug report to http://bugs.debian.org/ and include the crash backtrace, preprocessed source, and associated run script.
clang: note: diagnostic msg: 
PLEASE ATTACH THE FOLLOWING FILES TO THE BUG REPORT:
Preprocessed source(s) and associated run script(s) are located at:
clang: note: diagnostic msg: /tmp/EqualComparison-835ea0.cpp
clang: note: diagnostic msg: /tmp/EqualComparison-835ea0.sh
clang: note: diagnostic msg:

navsan:
The compiler is probably running out of memory. Can you try turning down
the level of parallelism?

pateljm:
LGTM. Merging. Thanks @saketj.

zuyu:
@saketj Next time, please squash all the commits into one before merging. Thanks!

saketj:
@zuyu Thanks, sure will do that.

PR #102, opened 2016-09-24
Refactored RebuildStatus.

zuyu:
Assigned to @hbdeshmukh. Thanks.

This PR wraps the rebuild status into a dedicate struct. It also allows future extension info for the distributed versions.

### DIFF
--- a/query_execution/QueryExecutionState.hpp
+++ b/query_execution/QueryExecutionState.hpp
@@ -92,8 +92,15 @@ class QueryExecutionState {
                                const std::size_t num_rebuild_workorders,
                                const bool rebuild_initiated) {
     DCHECK(operator_index < num_operators_);
-    rebuild_status_[operator_index].second = num_rebuild_workorders;
-    rebuild_status_[operator_index].first = rebuild_initiated;
+    auto search_res = rebuild_status_.find(operator_index);
+    if (search_res != rebuild_status_.end()) {
+      search_res->second.has_initiated = rebuild_initiated;
+      search_res->second.num_pending_workorders = num_rebuild_workorders;
+    } else {
+      RebuildStatus rebuild_status(rebuild_initiated, num_rebuild_workorders);
+
+      rebuild_status_.emplace(operator_index, std::move(rebuild_status));
+    }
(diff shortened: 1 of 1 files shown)


query_execution/QueryExecutionState.hpp
```diff
@@ -92,8 +92,15 @@ class QueryExecutionState {
...
-    rebuild_status_[operator_index].second = num_rebuild_workorders;
-    rebuild_status_[operator_index].first = rebuild_initiated;
+    auto search_res = rebuild_status_.find(operator_index);
+    if (search_res != rebuild_status_.end()) {
+      search_res->second.has_initiated = rebuild_initiated;
+      search_res->second.num_pending_workorders = num_rebuild_workorders;
+    } else {
+      RebuildStatus rebuild_status(rebuild_initiated, num_rebuild_workorders);
+
+      rebuild_status_.emplace(operator_index, std::move(rebuild_status));
```
hbdeshmukh: I am trying to understand the difference between the ``std::move`` usage here and if you use ``std::foward_as_tuple`` instead. For instance, line 100 can be removed and instead line 102 will look as follows:

``rebuild_status_.emplace(operator_index, std::std::forward_as_tuple(rebuild_initiated, num_rebuild_workorders));``

Do you think any one version is better than another?

hbdeshmukh:
Hi @zuyu 

This looks good to me except a small clarifying question.

query_execution/QueryExecutionState.hpp
zuyu: If we try to use `forward_as_tuple` (which produces a rvalue expression), it actually looks like as following:

```
      rebuild_status_.emplace(std::piecewise_construct,
                              std::forward_as_tuple(operator_index),
                              std::forward_as_tuple(rebuild_initiated, num_rebuild_workorders));
```

But I personally prefer the existing version, as it is more straightforward.

hbdeshmukh: Sounds good.

hbdeshmukh:
LGTM, merging.

PR #108, opened 2016-10-05
Remove unnecessary code from FastSeparateChainingHashTable::resize()

hbdeshmukh:
- In resize, all the values are moved from original hash table are
  copied to the new hash table using memcpy().
- Removed the for loop where values are constructed using 'new' operator, as
  all the values are trivially constructible in the specialized
  aggregation hash table implementation.


pateljm:
LGTM. Merging since Travis was run before.

PR #109, opened 2016-10-05
Refactored SplitRowStore bulk insertion

cramja:
The inner loop of the insert algorithm has been changed to reduce function calls to only those that are absolutely necessary. Also, we merge copies which come from other rowstore source, speeding up insertion time. 

Also adds support for the idea of 'partial inserts'. Partial inserts are when you are only inserting a subset of the columns at a time. Partial inserts will be used in a later commit.

*Testing*
Unit tests have been updated. The old bulkInsert tests needed to be modified because now we have situations where a block will not be filled up completely- only to a threshold value. This reduces the runtime of the costly inner loop at the cost of a few tuples.

*Performance*
I had a [similar PR-100 open](https://github.com/apache/incubator-quickstep/pull/100) last week. I ran TPCH SF100 queries 1-17 with this branch and with the branch from PR-100. They performed within a 1% margin of each other so it is safe to say that this branch is as fast as the last branch (which was 2x the base).

**Best times for contrived 1gb relation, 50% selectivity selection test**
(See PR100)
master: 1602.358 ms
splitrow_insert_refactor: 469.664 ms

**TPCH runtime Queries 1-16 (sum of the average of middle 3 runs of 5 runs total per query**
master: 10.61 minutes
splitrow_insert_refactor: 10.48 minutes

**TLDR** this is an incremental improvement.

### DIFF
--- a/utility/BitVector.hpp
+++ b/utility/BitVector.hpp
@@ -192,6 +192,7 @@ class BitVector {
    *            as this BitVector.
    **/
   inline void setMemory(void *ptr) {
+    DCHECK(!owned_);
     this->data_array_ = static_cast<std::size_t*>(ptr);
   }
 
--- a/utility/BitVector.hpp
+++ b/utility/BitVector.hpp
@@ -183,6 +183,19 @@ class BitVector {
   }
 
   /**
+   * @brief Assign this BitVector's contents to the pointed-to memory.
+   *
+   * @warning caller is responsible for ensuring the Bitvector has the correct
+   *          ownership and size.
+   *
+   * @param ptr Pointer to data representing a BitVector with the same parameters
+   *            as this BitVector.
+   **/
+  inline void setMemory(void *ptr) {
+    this->data_array_ = static_cast<std::size_t*>(ptr);
+  }
+
+  /**
    * @brief Similar to assignFrom(), but the other BitVector to assign from is
    *        allowed to be longer than this one.
    * @warning Only available when enable_short_version is false.
--- a/storage/SplitRowStoreTupleStorageSubBlock.cpp
+++ b/storage/SplitRowStoreTupleStorageSubBlock.cpp
@@ -41,54 +41,61 @@ namespace quickstep {
 
 QUICKSTEP_REGISTER_TUPLE_STORE(SplitRowStoreTupleStorageSubBlock, SPLIT_ROW_STORE);
 
+using splitrow_internal::CopyGroupList;
+using splitrow_internal::ContiguousAttrs;
+using splitrow_internal::NullableAttr;
+using splitrow_internal::VarLenAttr;
+
+const std::size_t SplitRowStoreTupleStorageSubBlock::kVarLenSlotSize = sizeof(std::uint32_t) * 2;
+
 namespace {
 
-template <typename ValueAccessorT, bool nullable_attrs>
-inline std::size_t CalculateVariableSize(
+  template<typename ValueAccessorT, bool nullable_attrs>
+  inline std::size_t CalculateVariableSize(
     const CatalogRelationSchema &relation,
     const ValueAccessorT &accessor) {
-  std::size_t total_size = 0;
-  attribute_id accessor_attr_id = 0;
-  for (CatalogRelationSchema::const_iterator attr_it = relation.begin();
-       attr_it != relation.end();
-       ++attr_it, ++accessor_attr_id) {
-    if (!attr_it->getType().isVariableLength()) {
-      continue;
-    }
+    std::size_t total_size = 0;
+    attribute_id accessor_attr_id = 0;
+    for (CatalogRelationSchema::const_iterator attr_it = relation.begin();
+         attr_it != relation.end();
+         ++attr_it, ++accessor_attr_id) {
+      if (!attr_it->getType().isVariableLength()) {
+        continue;
+      }
 
-    TypedValue value(accessor.getTypedValue(accessor_attr_id));
-    if (nullable_attrs && value.isNull()) {
-      continue;
+      TypedValue value(accessor.getTypedValue(accessor_attr_id));
+      if (nullable_attrs && value.isNull()) {
+        continue;
+      }
+      total_size += value.getDataSize();
     }
-    total_size += value.getDataSize();
+    return total_size;
   }
-  return total_size;
-}
 
-template <typename ValueAccessorT, bool nullable_attrs>
-inline std::size_t CalculateVariableSizeWithRemappedAttributes(
+  template<typename ValueAccessorT, bool nullable_attrs>
+  inline std::size_t CalculateVariableSizeWithRemappedAttributes(
     const CatalogRelationSchema &relation,
     const ValueAccessorT &accessor,
     const std::vector<attribute_id> &attribute_map) {
-  std::size_t total_size = 0;
-  std::vector<attribute_id>::const_iterator attr_map_it = attribute_map.begin();
-  for (CatalogRelationSchema::const_iterator attr_it = relation.begin();
-       attr_it != relation.end();
-       ++attr_it, ++attr_map_it) {
-    if (!attr_it->getType().isVariableLength()) {
-      continue;
-    }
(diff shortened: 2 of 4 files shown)


cramja:
Note: this crashes GCC5. I'm not sure what to do about this. Maybe @navsan or @hakanmemisoglu (hakan the compiler man) might have some advice.

```
<lambda(auto:1*)>]’
/fastdisk/quickstep/storage/ValueAccessorUtil.hpp:263:55:   required from ‘auto quickstep::InvokeOnAnyValueAccessor(quickstep::ValueAccessor*, const FunctorT&) [with FunctorT = quickstep::SplitRowStoreTupleStorageSubBlock::bulkInsertPartialTuplesImpl(const quickstep::splitrow_internal::CopyGroupList&, quickstep::ValueAccessor*, std::size_t) [with bool copy_nulls = true; bool copy_varlen = true; bool fill_to_capacity = true; quickstep::tuple_id = int; std::size_t = long unsigned int]::<lambda(auto:1*)>]’
/fastdisk/quickstep/storage/SplitRowStoreTupleStorageSubBlock.cpp:342:27:   required from ‘quickstep::tuple_id quickstep::SplitRowStoreTupleStorageSubBlock::bulkInsertPartialTuplesImpl(const quickstep::splitrow_internal::CopyGroupList&, quickstep::ValueAccessor*, std::size_t) [with bool copy_nulls = true; bool copy_varlen = true; bool fill_to_capacity = true; quickstep::tuple_id = int; std::size_t = long unsigned int]’
/fastdisk/quickstep/storage/SplitRowStoreTupleStorageSubBlock.cpp:278:94:   required from here
/fastdisk/quickstep/storage/SplitRowStoreTupleStorageSubBlock.cpp:325:68: internal compiler error: in tsubst_copy, at cp/pt.c:13040
   const std::size_t num_c_attr = copy_groups.contiguous_attrs_.size();
                                                                    ^
Please submit a full bug report,
with preprocessed source if appropriate.
See <file:///usr/share/doc/gcc-5/README.Bugs> for instructions.

```

hbdeshmukh:
@cramja Do you mean TPC-H SF100 Q1-Q17 has 2x improvement?

navsan:
Have you tried with a newer or older version of GCC5?

hakanmemisoglu:
Hi @cramja, the problem might be related lambda capture arguments. GCC 5 was giving the same problems if you use general capture by reference [&]. You can try to give specific reference parameters that is used in lambda function, such as [&var1, &var2]. It should fix the problem.

navsan:
This bug was filed and fixed a year ago in 5.3, I think. It had been around for 2-3 years, so we’ll probably be affected in older versions of GCC too. 
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=67411

Any chance you can simply change the code to do something equivalent by syntactically different? Maybe that’ll help?

cramja:
@hbdeshmukh 2x improvement was for the contrived test of trying to insert a whole bunch of tuples into a Splitrow at once as in PR100. As for TPCH, I'm not sure, other than that this is not any slower. I could run those tests again, but right now I'm working on the compiler issue.

cramja:
Thanks @hakanmemisoglu I tried something similar to what you were suggesting and the "works but I don't know why" solution I found was to declare the variables inside the lambda. It's ever so slightly less efficient code but w/e, compiles.

cramja:
@hbdeshmukh do we have any recent TPCH 100 results from a CloudLab box?

cramja:
@hbdeshmukh I updated for TPCH in the PR header

utility/BitVector.hpp
```diff
@@ -183,6 +183,19 @@ class BitVector {
...
+   * @brief Assign this BitVector's contents to the pointed-to memory.
+   *
+   * @warning caller is responsible for ensuring the Bitvector has the correct
+   *          ownership and size.
+   *
+   * @param ptr Pointer to data representing a BitVector with the same parameters
+   *            as this BitVector.
+   **/
+  inline void setMemory(void *ptr) {
+    this->data_array_ = static_cast<std::size_t*>(ptr);
```
navsan: Doesn't this potentially lead to a memory leak?

cramja: Yes, it could if the `BitVector` owned memory and then is reassigned via this method. 

This certainly is a dangerous method but you can see why I used it right? Instead of having to instantiate an entirely new BitVector I essentially want to use BitVector's methods over a window of memory.

Adding a `DCHECK(owned)` here should demonstrate to programmers the proper usage.

cramja:
@navsan Ready to merge?

cramja:
@jianqiao any issues?

navsan:
Have you been able to test this on TPCH and make sure that the results are
correct? If so we can go ahead and merge. I haven't been able to test this
for certain queries yet.

cramja:
If that's what were waiting on, then yes, I can go test it. Of course that will be with a subset of working queries.

jianqiao:
@cramja In general looks good. I agree with Navneet that `setMemory()` does not look like a safe public method (there are other `const` members in the `BitVector` class not addressed) -- the method seems to be just applicable in very limited scenarios. Anyway let's merge this PR first, and I can help revise `setMemory()` later when working on the reordering-output-attributes stuff.

Can you rebase the branch and commit it to `apache:splitrow_insert_refactor`? Seems that I cannot access the `cramja:splitrow_insert_refactor` branch.

cramja:
@jianqiao Sure, that sounds good. @navsan and I talked about another alternative which is to call the BitVector constructor instead of `setMemory`. I made the `set` method because I thought that it would be cheaper than instantiation of an object, but navneet explained that since the compiler would pre allocate space on the stack, that the costs would be equivalent. Do you think that would be a good alternative?

jianqiao:
@cramja I'm not sure about the overhead for calling the constructor inside the accessor loop. We can first have this `setMemory()` version merged to have a reference for the performance.

cramja:
I ran TPCH 10 on a cloudlab instance. The update is marginally better. Note that *master* is using splitrows, not columnstores. Also this is queries `01 04 06 09 10 11 12 13 14 15 16 19 22`

|master|splitrow refactor| master - splitrow refactor|
|---|---|---|
|   2244.56|   2227.01|      17.5|
|    482.29|    475.22|       7.1|
|    298.32|    295.05|       3.3|
|    823.28|    822.34|       0.9|
|   1655.41|   1680.74|     -25.3|
|    343.62|    360.24|     -16.6|
|    566.85|    568.88|      -2.0|
|   3189.44|   3104.58|      84.9|
|    315.47|    312.61|       2.9|
|    594.03|    555.04|      39.0|
|   1076.28|   1132.68|     -56.4|
|    448.39|    441.66|       6.7|
|    776.47|    777.97|      -1.5|
|  12814   |12754|60.39|
**Note the last row is a total**


Comparing it against *Master* on which uses compressed columnstores, I found:

|master_cs|splitrow refactor| master_cs - splitrow refactor|
|---|---|---|
|   1676.65|   2227.01|    -550.4|
|    324.30|    475.22|    -150.9|
|     44.78|    295.05|    -250.3|
|    692.21|    822.34|    -130.1|
|   1522.79|   1680.74|    -158.0|
|    325.94|    360.24|     -34.3|
|    210.29|    568.88|    -358.6|
|   3278.79|   3104.58|     174.2|
|     98.74|    312.61|    -213.9|
|    430.47|    555.04|    -124.6|
|    977.81|   1132.68|    -154.9|
|    113.18|    441.66|    -328.5|
|    779.42|    777.97|       1.5|
|10475.37|12754.02|-2278.64|

**Note the last row is a total**

So apparently compression is the way to go here. Though, this is of course still an improvement on the old splitrow.

jianqiao:
Merged and closed.

pateljm:
@jianqiao Did you merge this? The PR is still open. 

@cramja Thanks for doing this. Nice!

jianqiao:
@pateljm The PR was merged but was not closed automatically, @cramja can manually close it.

cramja:
@jianqiao thanks! will close.

PR #112, opened 2016-10-10
Minor bug fix in AggregationOperationState

hbdeshmukh:
Replace getHashTable() call with getHashTableFast() call.

Thanks @jianqiao for pointing out the bug.


jianqiao:
LGTM. Merging.

PR #115, opened 2016-10-18
Update travis to run only single thread while building release version

saketj:
This a one-line change PR that fixes Travis build job failures due to compiler out-of-memory errors by running a single thread while building a release version.


jianqiao:
LGTM!

PR #119, opened 2016-10-20
Fix a bug in FastSeparateChainingHashTable::destroy payload

hbdeshmukh:
- Change in pointer arithmetic now ensures that destroyPayload will be
  called for correct payload entry in each bucket.

Thanks @jianqiao in fixing this bug.


jianqiao:
LGTM! Merging.

PR #120, opened 2016-10-21
Store handle count in FastSeparateChainingHashTable

hbdeshmukh:
- Using handles_ vector directly in FastSeparateChainingHashTable can
  cause problems. One example is a distinctify hash table which doesn't
  accept any payload. In this case, the reference to handles_ vector
  becomes garbage after the constructor gets executed. Therefore to
  avoid faults about accessing an invalid reference, we use the number
  of handles as a guarding mechanism.

- Verified by reducing the hash table size artificially, so as to forcing a ``resize()`` operation. The ``destroyPayload()`` function gets called in ``resize()``. The fix in this PR fixes the behavior of ``destroyPayload`` function.

### DIFF
--- a/storage/FastSeparateChainingHashTable.hpp
+++ b/storage/FastSeparateChainingHashTable.hpp
@@ -145,10 +145,10 @@ class FastSeparateChainingHashTable
         header_->buckets_allocated.load(std::memory_order_relaxed);
     void *bucket_ptr = static_cast<char *>(buckets_) + kValueOffset;
     for (std::size_t bucket_num = 0; bucket_num < num_buckets; ++bucket_num) {
-      for (std::size_t handle_num = 0; handle_num < handles_.size(); ++handle_num) {
+      for (std::size_t handle_id = 0; handle_id < num_handles_; ++handle_id) {
         void *value_internal_ptr =
-            static_cast<char *>(bucket_ptr) + this->payload_offsets_[handle_num];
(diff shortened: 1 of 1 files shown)


hbdeshmukh:
Hi @hakanmemisoglu 

Can you please review this PR?

storage/FastSeparateChainingHashTable.hpp
```diff
@@ -145,10 +145,10 @@ class FastSeparateChainingHashTable
         header_->buckets_allocated.load(std::memory_order_relaxed);
     void *bucket_ptr = static_cast<char *>(buckets_) + kValueOffset;
     for (std::size_t bucket_num = 0; bucket_num < num_buckets; ++bucket_num) {
-      for (std::size_t handle_num = 0; handle_num < handles_.size(); ++handle_num) {
+      for (std::size_t handle_id = 0; handle_id < num_handles_; ++handle_id) {
         void *value_internal_ptr =
```
hakanmemisoglu: Hi @hbdeshmukh,
It is not critical but I have a question.
Is there a particular reason for `value_internal_ptr` 's type is `void *` instead of `char *`. Right hand side of the assignment has type of `char *`.

jianqiao: I think it's okay either way. If we are using `char *`, then we need `reinterpret_cast` to cast `char *` to `std::uint8_t *`. Here two `static_cast`'s should semantically be better.

jianqiao:
LGTM. Merging.

PR #125, opened 2016-11-01
QUICKSTEP-61 Fixes varlen insert bug.

cramja:
This change fixes an incorrect estimate for the maximum size of a tuple. Previously, it was not accounting for the space used by the fixed length portion of the tuple.

The unit tests were modified because, now, we insert fewer tuples. It's actually quite hard to test the correct number of tuples which will be inserted, so this has been left as a TODO. We were able to load all of the SSB tables correctly now.


cramja:
@hbdeshmukh Here is the change

PR #128, opened 2016-11-06
Refine estimates for estimateCardinality for TableReference

hbdeshmukh:
- Use exact CatalogRelation statistics in SimpleCostModel's
  estimateCardinality method, whenever stats on that relation are available.

This feature particularly helps in TPC-H Q21 (100 scale factor), in which there are two large hash tables built on the lineitem (cardinality: 600M) table. Such large hash tables leave very little room in the buffer pool for other operators, causing evictions and resulting in a large execution time. 

In the original branch, the estimate for the number of entries in the hash table, as provided by the cost model is 1.09B. After this PR, we can reduce that estimate to 600M. Running Q21 the CloudLab machine, gave the following results: (time in milliseconds).

| Original | After changes | Speed-up |
|----------|---------------|----------|
| 178,787  | 12,043        | 14.85    |


jianqiao:
LGTM. Merging.

PR #130, opened 2016-11-08
Allows filters to be evaluated in SMA

cramja:
On re-running SSB100, I found that we now have a situation where filters are used against SMAs. This is actually very easy to handle: either return the filter, or don't return the filter if your selectivity is ALL/NONE.

@hbdeshmukh I will update the SSB numbers for SMA here when this has finished compiling/running.


cramja:
@hbdeshmukh can I get a +1 so we can merge/close this?

hbdeshmukh:
Looks good to me, merging.

PR #131, opened 2016-11-13
Unified WorkOrderCompletionMessage proto.

zuyu:
Assigned to @hbdeshmukh.

This PR combined `WorkOrderCompletionMessage` for both normal and rebuild work orders. And it is required for the distributed version to refactor `PolicyEnforcerBase::decrementNumQueuedWorkOrders` to use `WorkOrderCompletionMessage` as the argument instead.

### DIFF
--- a/query_execution/PolicyEnforcerBase.cpp
+++ b/query_execution/PolicyEnforcerBase.cpp
@@ -45,7 +45,7 @@ void PolicyEnforcerBase::processMessage(const TaggedMessage &tagged_message) {
 
   switch (tagged_message.message_type()) {
     case kWorkOrderCompleteMessage: {
-      serialization::NormalWorkOrderCompletionMessage proto;
+      serialization::WorkOrderCompletionMessage proto;
       // Note: This proto message contains the time it took to execute the
       // WorkOrder. It can be accessed in this scope.
       CHECK(proto.ParseFromArray(tagged_message.message(),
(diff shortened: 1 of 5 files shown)


hbdeshmukh:
Hi Zuyu,

We had a discussion about this design earlier and we came up to the conclusion that it serves us better in the future to have different types for normal work order completion and rebuild work order completion. In fact, I also documented the reasons for this bifurcation in the source code. For convenience, I am describing that comment below: 

```
Note: There are different types of completion messages for normal work orders  
rebuild work orders. This can be potentially helpful when we want to collect   
different statistics for executing different types of work orders.                
e.g. In select normal work order completion message, we could be interested       
in the selectivity of the block whose work order got execute. In rebuild work  
order completion message, we may be interested in adding the compression          
ratio or dictionary size of the rebuilt block.
```

Given this, I am not sure if we still need to go back to the old way.

zuyu:
Yes, and if there is a new field dedicated to one case, say `RebuildWorkOrder`, we would create the new protobuf message extending `WorkOrderCompletionMessage` as following:

```
message RebuildWorkOrderCompletionMessage {
  extends WorkOrderCompletionMessage {
    // Add the new field.
  }
}
```

As I said in the PR description, I need this way in the following PR #137 to refactor [`PolicyEnforcerBase::decrementNumQueuedWorkOrders`](https://github.com/apache/incubator-quickstep/pull/137/files#diff-fa94240de06b90bf57c8b9fdc3403d71R93).

And I welcome alternative ways to achieve the same goal as #137.

zuyu:
@hbdeshmukh Note that beside your quote, there is also another related quote after:

> // TODO(harshad) : If there are different fields in the two message types below,
// create a base message class called WorkOrderCompletionMessage and make the
// two classes below extend the base class. All the common fields in both the
// classes can be moved to the base class.

hbdeshmukh:
Alright, thanks for the clarification. Looks good to me.

hbdeshmukh:
Because lot of branches got merged meanwhile, can you please rebase with master and let me know? I will merge this branch after that.

zuyu:
@hbdeshmukh Thanks for the review. I have synced with the master, and it is ready to merge. Cheers!

PR #134, opened 2016-11-15
Removed StorageManager from QueryProcessor.

zuyu:
This small PR removed the `StorageManager` dependency in `QueryProcessor`.


hbdeshmukh:
Can you please explain why is the dependency unnecessary?

zuyu:
For now, the only place uses `QueryProcessor` is the `QuickstepCLI`, and `StorageManager` is not required in the optimizer. On the other hand, in the distributed version, `Conductor` will re-use `QueryProcessor`, but does not need `StorageManger` at all.

So I think it is better to have both `StorageManager` and `QueryProcessor` in the `QuickstepCLI`.

hbdeshmukh:
Looks good to me, merging.

PR #136, opened 2016-11-15
Moved some storage-related constants.

zuyu:
This small PR moved two storage-related constants into a header file so that other files in the distributed version would also use them.


hbdeshmukh:
Looks good to me.

PR #139, opened 2016-11-17
Remove packedrow

cramja:
When I refactored the SplitRowStore to handle partial inserts, I optimized the bulk insert algorithm to efficiently handle inserts where attributes are all fixed length. There is almost no difference between the runtime of SSB/TPCH with/without the packed row store. So, in the interest of keeping simple code, I've removed the now unnecessary `splitrowstore`.


cramja:
@hbdeshmukh my new years resolution is to finish this PR. Want to review this with me?

hbdeshmukh:
Hi @cramja - Sure, we can close this PR. Looks like Travis is failing right now and the issue is not the OOM error we saw in some of the previous PRs. Once you fix the issues, please let me know. I can take a look.

cramja:
@hbdeshmukh I fixed the problem, hopefully. Waiting for Travis to come back. 


There's several Travis runs which I triggered (`git commit --amend; git push --force`) while testing. Did travis remove the kill build feature? I seem to remember being able to kill these sorts of things in the past by logging into their site...

hbdeshmukh:
Looks good to me.

PR #141, opened 2016-11-17
Refactor WorkOrderFactory for a better debug info.

zuyu:


cramja:
LGTM. 

I image debugging the distributed version is difficult and I am guessing that is what this is related to. Was this change warranted by anything specific?

zuyu:
I guess by `warrant` you mean whether it affects others and its correctness.

It is only related to the distributed version and `glog` output.

cramja:
I meant, was the reason you added it because of anything specific you were having trouble with? But still, LGTM.

zuyu:
Yes, it helped me debug the distributed query execution for queries with hash-joins and aggregations so that I know which `WorkOrder` executes on which `Shiftboss`.

PR #142, opened 2016-11-18
Distributed Multi-shiftboss Query Execution Unit Test

zuyu:
Do not merge until #137 is in.

This PR allows the distributed unit test to use multi-Shiftboss so that it simulates the full-functional distributed query execution engine.

Note that this naive `WorkOrder` scheduling implementation uses a simple round-robin manner.


zuyu:
Need to split into multiple PRs.

PR #148, opened 2016-11-20
Transferred the ownership of QueryHandle to QueryManager.

zuyu:
This PR allows the `QueryManager` to take over `QueryHandle` during the query execution, thus it is easier for the distributed version to execute the concurrent queries w/o worrying about the ownership of each `QueryHandle`.

### DIFF
--- a/cli/CommandExecutor.cpp
+++ b/cli/CommandExecutor.cpp
@@ -213,25 +213,27 @@ inline std::vector<TypedValue> executeQueryForSingleRow(
   DCHECK(result.condition == ParseResult::kSuccess);
 
   const ParseStatement &statement = *result.parsed_statement;
+  const CatalogRelation *query_result_relation = nullptr;
 
-  // Generate the query plan.
-  std::unique_ptr<QueryHandle> query_handle(
-      std::make_unique<QueryHandle>(query_processor->query_id(),
-                                    main_thread_client_id,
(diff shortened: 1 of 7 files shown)


hbdeshmukh:
Hi @zuyu and @pateljm 

This change looks good to me and I can handle the overlap of the code in this branch with the code for concurrent execution. 

@zuyu Can you please rebase this branch with master, enable the DEBUG build in Travis and let me know once it comes back green? Then I will merge this branch.

zuyu:
Hi @hbdeshmukh,

This PR has synced with the master.

But I did not get the reason why we need to try the debug mode build in Travis CI. Could you please explain more? Thanks!

hbdeshmukh:
Hi @zuyu 

As I mentioned on the dev list discussion, we can turn on those DEBUG build configurations which don't time out. That way we will have some assurance about DEBUG ASSERT statements being passed.

zuyu:
Hi @hbdeshmukh,

For the same source code with some bugs inside, does the debug build produce the different CI results than the release build?

I don't think so, and that it is why I think the debug build and `DEBUG ASSERT` do not matter for here for code reviews.

zuyu:
Hi @hbdeshmukh,

Could you please take a look again, as the debug build using `LLVM` also passed?

PR #154, opened 2016-12-15
Quickstep-63 Fix PlanVisualizer's crashes on physical plans unsupported by StarSchemaSimpleCostModel.

jianqiao:
`PlanVisualizer` would result in crashes on some queries (i.e., `CreateTable`) which do not have corresponding physical plan support in `StarSchemaSimpleCostModel::estimateCardinality`.

This PR fixes this problem by:
(1) Let the cost models throw `UnsupportedPhysicalPlan` exception for unsupported plans.
(2) `PlanVisualizer` simply catches the exceptions and omits the cardinality info for unsupported physical nodes.

Assigned to @zuyu


zuyu:
LGTM.

PR #155, opened 2016-12-17
Add min/max statistics and the exactness flag into CatalogRelationStatistic

jianqiao:
This PR collects the min/max statistics for each column of a table with the `\analyze` command. It also adds an `is_exact_` flag into `CatalogRelationStatistic` to indicate whether the statistics reflect the exact (i.e. not approximate) properties of the current data. Each table's `is_exact_` flag will be set to `true` with the `\analyze` command on that table, and will be set to `false` with the  _insert/delete/update/copy_ operations on that table.

This PR is a preparation for implementing some fast-path operators (e.g. exact filters and vector-based aggregation) that significantly improve Quickstep's performance on some of the SSB/TPC-H queries.

**NOTE**: With this PR the system will be incompatible with previously created databases due to some changes in `Catalog.proto`. In fact, we have two options:
(1) Make it backward compatible. The downside of this option is that there will be some fields left "obsolete" in `Catalog.proto`. I.e. the fields remain there but will not be used in the future by the system.
(2) Make the changes without backward compatiblity and recreate+reload the databases.

To me, option (2) seems affordable at this stage of Quickstep, and CatalogRelationStatistic is not likely to incur any future backward-incompatible catalog schema changes.

PR #156, opened 2017-01-06
Fixed the distributed unit test failures in resetting exactness for InsertSelect.

zuyu:
Fixed the distributed unit test failures introduced in #155.

Assigned to @jianqiao.


jianqiao:
LGTM. Thanks for the fix! Will merge after travis returns green.

PR #159, opened 2017-01-12
Refactored SelectOperator::getAllWorkOrders.

zuyu:
Assigned to @hbdeshmukh.

This PR simplified `SelectOperator::getAllWorkOrders`, and allows to generate `WorkOrder`s when the input relation has data partitions, but not in the `NUMA` case.


hbdeshmukh:
Looks good to me.

PR #163, opened 2017-01-18
Marked LIP as a non-default argument.

zuyu:
Assigned to @jianqiao.


jianqiao:
LGTM. Is there some special reason for doing this change?

zuyu:
@jianqiao First, it is doable. Second, I will add a argument with default value, without worrying about the argument order in terms of coding style.

PR #165, opened 2017-01-20
Added Operator support for Partitioned HashJoin.

zuyu:
Assigned to @hbdeshmukh or @jianqiao.

This PR allows `HashJoin` to accept partitioned info. If no partition, `num_partitions` is `1`.

### DIFF
--- a/relational_operators/HashJoinOperator.cpp
+++ b/relational_operators/HashJoinOperator.cpp
@@ -207,49 +207,43 @@ bool HashJoinOperator::getAllNonOuterJoinWorkOrders(
         query_context->getScalarGroup(selection_index_);
     InsertDestination *output_destination =
         query_context->getInsertDestination(output_destination_index_);
-    const JoinHashTable &hash_table =
-        *(query_context->getJoinHashTable(hash_table_index_));
 
     if (probe_relation_is_stored_) {
-      if (!started_) {
-        for (const block_id probe_block_id : probe_relation_block_ids_) {
+      if (started_) {
+        return true;
+      }
+
+      for (std::size_t part_id = 0; part_id < num_partitions_; ++part_id) {
+        const JoinHashTable &hash_table =
+            *(query_context->getJoinHashTable(hash_table_index_, part_id));
+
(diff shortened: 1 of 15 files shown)


storage/StorageBlockInfo.hpp
```diff
@@ -49,6 +50,8 @@ static constexpr int kBlockIdDomainLengthInDigits = std::numeric_limits<block_id
 static constexpr int kBlockIdCounterLengthInDigits = std::numeric_limits<block_id_counter>::digits10;
 static constexpr block_id_domain kMaxDomain = static_cast<block_id_domain>(0xFFFF);
 
+typedef std::vector<block_id> blocks_in_partition;
```
jianqiao: `block_in_partition` is not a primitive type. Perhaps use `BlockInPartition`?

relational_operators/HashJoinOperator.cpp
```diff
@@ -207,49 +207,43 @@ bool HashJoinOperator::getAllNonOuterJoinWorkOrders(
...
-                residual_predicate,
-                selection,
-                hash_table,
-                output_destination,
-                storage_manager,
-                CreateLIPFilterAdaptiveProberHelper(lip_deployment_index_, query_context)),
-            op_index_);
-        ++num_workorders_generated_;
-      }  // end while
+      for (std::size_t part_id = 0; part_id < num_partitions_; ++part_id) {
```
jianqiao: How many partitions would we expect to have in typical situations?

zuyu: It depends on the input relation and the optimizer. The former may have given partitions (i.e., 4), and the latter may add an explicit partition phase for a relation w/o partitions.

I do so to minimize the code changes without hurting the performance much.

jianqiao: I see. It should be good.

storage/StorageBlockInfo.hpp
zuyu: Done.

jianqiao:
LGTM. Merging.

PR #166, opened 2017-01-22
Make the third party directory leaner

hbdeshmukh:
- Added shell script to download prerequisite third party libs, instead
  of relying on pre-existing source files in the third party directory.
- Deleted third party library code for some modules.
- Added re2 CMakeLists patch file.
- Ignore warnings during re2 compilation.
- Compilation fixes due to google benchmark upgrade.
- Fixes for the compilation error in Window aggregation unit test.
- Perform check for the current directory before running the third party
  scripts.
- Created macro for setting GFLAG library name.
- Move third party source files to third_party/src directory.
- Update travis workflow.


zuyu:
Hi @hbdeshmukh,

I was wondering why we download release directly, instead of using submodules for `re2` and `gtest`? And, do we need to run the download shell script manually, or is it a part of `make` after generating `cmake` configurations?

hbdeshmukh:
Hi @zuyu 

There was an elaborate discussion on the dev-list about this branch. Can you please refer to that instead of me reproducing it here? Thanks.

hbdeshmukh:
We need to run the script manually. I am not aware of a way through which make build process can do this.

zuyu:
I found two mail threads regarding this topic: `release: third_party/` and `Fixing git submodule commits to a release`. Do I still miss something, as I did not get the point of not using submodules?

zuyu:
`cmake`'s [execute_process](https://cmake.org/cmake/help/latest/command/execute_process.html) would run a given script for you.

hbdeshmukh:
Combining CMake build process and downloading of third party libraries is not desirable. Here's the proposed workflow in the new branch: 

1. Run download third party script (that also applies patches)
2. CMake (add subdirectories for third party libs and invoke Quickstep's CMake files)
3. make (build everything)

Note that 1 needs to be run only once. However 2 and 3 will be run over and over again as we continue development. If 1 and 2 are combined, we will repeatedly download third party libs and apply patches unnecessarily.

zuyu:
I understand the proposal, and my suggestion is to avoid the extra step 1. There should have a way in step 3 to run once the download script and apply patches (i.e., check for whether a particular file created by the script exists), while building files as many as the dev likes.

CMakeLists.txt
```diff
@@ -694,8 +698,9 @@ else()
   include_directories(${CMAKE_CURRENT_BINARY_DIR}/third_party)
 endif()
 
-add_subdirectory ("${THIRD_PARTY_SOURCE_DIR}/re2_cmake" "${CMAKE_CURRENT_BINARY_DIR}/third_party/re2")
+#add_subdirectory ("${THIRD_PARTY_SOURCE_DIR}/re2_cmake" "${CMAKE_CURRENT_BINARY_DIR}/third_party/re2")
 include_directories("${THIRD_PARTY_SOURCE_DIR}/re2")
+add_subdirectory ("${THIRD_PARTY_SOURCE_DIR}/re2" "${CMAKE_CURRENT_BINARY_DIR}/third_party/re2")
```
zuyu: Per the conventional usage in this file, please move this `add_subdirectory` before `include_directories`, while deleting the old `add_subdirectory`.

hbdeshmukh: Removed the dead code. If I change the order of 702 and 703, it gives me following error, and hence include first and add subdirectory later. 
```
...third_party/src/re2/re2/bitstate.cc:24:26: fatal error: util/logging.h: No such file or directory
 #include "util/logging.h"
```

hbdeshmukh:
>I understand the proposal, and my suggestion is to avoid the extra step 1. There should have a way in step 3 to run once the download script and apply patches (i.e., check for whether a particular file created by the script exists), while building files as many as the dev likes.

I am unable to understand your point. What's the problem with the download step that you have to perform only once in your development lifecycle? As I said earlier, making cmake and make build processes complicated is not desired.

zuyu:
Well, you need to be consistent in [the document](https://github.com/apache/incubator-quickstep/blob/master/README.md#giving-it-a-spin) regarding this extra step.

cramja:
@zuyu are you suggesting that the `cmake` command should automatically run the download script?

zuyu:
Yes.

cramja:
@zuyu I think that's a good idea. Maybe though, that could be a version 2 feature. Seeing as before we had 3 commands with `submodule`, this is an improvement.

cramja:
Also, @hbdeshmukh, I ran it on my machine, +1

cramja:
Though, I would say, one enhancement we can do right now, is check that the user has `curl` and stop the script if they don't.

hbdeshmukh:
> Though, I would say, one enhancement we can do right now, is check that the user has curl and stop the script if they don't.

I have added the checks for the existence of ``curl`` and ``tar``.

hbdeshmukh:
@zuyu Once this PR is merged, I will change the instructions for building Quickstep.

zuyu:
In the last two builds, one of GCC jobs failed due to the time out. Is there a workaround?

hbdeshmukh:
I am assuming you are talking about [build 724 for PR#167](https://travis-ci.org/apache/incubator-quickstep/builds/195394320) and [build 725 for this PR](https://travis-ci.org/apache/incubator-quickstep/builds/196001823). 

I am not sure why one specific gcc build is timing out. With regards to the changes in this PR, the download and unzip process finishes quickly (1-2 seconds), so the overhead isn't high, if that's what you are asking.

zuyu:
Hi @hbdeshmukh,

While I am testing your branch on my Mac, could you please sync with the master? Thanks!

hbdeshmukh:
Synced. Thanks for testing.

zuyu:
Hi @hbdeshmukh,

The `cmake` configure for the single node works, but it fails for the distributed version (assume both protobuf 3.0 and grpc are installed):

```
CMake Error at third_party/src/tmb/third_party/gflags/CMakeLists.txt:315 (add_dependencies):
  Cannot add target-level dependencies to alias target "gflags".
```

After checkouting your own fork and the branch `lean-third-party`, what I did was:
```
 $ git submodule update --init
 $ cd third_party
 $ ./download_and_patch_prerequisites.sh
 $ cd ../build
 $ cmake -DCMAKE_BUILD_TYPE=Debug -DENABLE_DISTRIBUTED=ON ..
```

zuyu:
Hi @hbdeshmukh,

It seems that `third_party/src/gflags` and `tmb/third_party/gflags` are different versions.

hbdeshmukh:
Why can't TMB use the gflags library from Quickstep's 3rd party libraries?

zuyu:
It should be possible. Do you know how to do that?

On the other hand, if tmb builds gflags as a shared library, it would work.

hbdeshmukh:
Hi @zuyu 

I will request you to figure out a solution for the distributed issue, as you know it much better than anyone else. Any solution that you can make it to work should be fine. We should close this PR soon, as it has blocked our release. Thanks for your time.

zuyu:
Hi @hbdeshmukh,

I have [a fix](https://github.com/apache/incubator-quickstep/pull/168/commits/66178d7e11543b0d4bac59481e4b7dfee034acd6) that downgrades `gflags` to 2.1.2, the version we currently use, from the latest. It also removed the trailing whitespace in the shell script.

hbdeshmukh:
Sounds good @zuyu. Can you please create a PR to merge your fix to my branch https://github.com/hbdeshmukh/incubator-quickstep/tree/lean-third-party ? I will keep your commit as a separate commit in the branch, and then we can merge this branch.

zuyu:
Hi @hbdeshmukh,

I think I don't have the permission to push to your fork, but I have the following suggestions for you to do the same thing in your local fork, assuming in the local clone of `hbdeshmukh/incubator-quickstep`, not the `asf` repo:

```
 $ git remote add zuyu https://github.com/zuyu/incubator-quickstep.git
 $ git fetch zuyu lean-third-party
 $ git push origin lean-third-party
```

Let me know if you have questions.

hbdeshmukh:
Done. Travis somehow determined that there was a previous build and therefore didn't trigger another build, I think.

zuyu:
LGTM. Merged.

PR #170, opened 2017-01-29
Add unit test for CatalogRelationStatistics

jianqiao:
This PR adds unit tests for CatalogRelationStatistics. It tests that the `\analyze` command can produce correct stats, and that the `is_exact_` flag be set properly with insert / update / delete / analyze operations.

### DIFF
--- a/catalog/tests/CatalogRelationStatistics_unittest.cpp
+++ b/catalog/tests/CatalogRelationStatistics_unittest.cpp
@@ -1,0 +1,219 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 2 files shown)


jianqiao:
Assigned to @zuyu.

catalog/tests/CatalogRelationStatistics_unittest.cpp
```diff
@@ -0,0 +1,214 @@
...
+  executeQuery("CREATE TABLE analyzetest(x INT, y DOUBLE);");
+  executeQuery("INSERT INTO analyzetest VALUES(0, -0.5);");
+  executeQuery("INSERT INTO analyzetest VALUES(1, 0);");
+  executeQuery("INSERT INTO analyzetest VALUES(0, 0.5);");
+  executeAnalyze("analyzetest");
+
+  const CatalogRelation *relation = getRelationByName("analyzetest");
+  const CatalogRelationStatistics &stat = relation->getStatistics();
+
+  EXPECT_EQ(stat.getNumTuples(), 3u);
```
zuyu: Code style: please swap `EXPECT_EQ`'s both sides (i.e., `EXPECT_EQ(3u, stat.getNumTuples());`) and all below, so that the left side is the expected value, and the right side is the checked value, per the conventions.

catalog/tests/CatalogRelationStatistics_unittest.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+  EXPECT_EQ(stat.getNumDistinctValues(0), 2u);
```
zuyu: I assume the magic number `0` for the `attribute_id`. Please define a const `kFirstAttributeId` in the anonymous namespace above instead. Same for `1` below.

jianqiao:
Updated. Now closing and reopening the PR to restart travis.

PR #171, opened 2017-01-29
QUICKSTEP-68 Reorder intermediate relations'  attributes to improve copy performance.

jianqiao:
Currently, all the intermediate relations (i.e. temporary relations created during query execution) have `SPLIT_ROW_STORE` layouts. One existing optimization that improves copy performance for `SPLIT_ROW_STORE` layout is that, if two (or more) attributes are consecutive in both the source relation and the destination relation, then we can memory-copy the attributes' byte stream with just one `std::memcpy` call.

This PR is a complementary improvement that it adjusts the ordering of the attributes in the physical plan to maximize the opportunity of copying consecutive attributes – i.e. to minimize the number of `std::memcpy` calls.

This PR uses a simple greedy algorithm to reorder the attributes, which works well with SSB/TPCH workloads. The algorithm may be further improved later.

The feature can be turned on/off (default is **on**) with the `-reorder_columns` flag.

### DIFF
--- a/query_optimizer/PhysicalGenerator.cpp
+++ b/query_optimizer/PhysicalGenerator.cpp
@@ -28,6 +28,7 @@
 #include "query_optimizer/physical/Physical.hpp"
 #include "query_optimizer/rules/AttachLIPFilters.hpp"
 #include "query_optimizer/rules/PruneColumns.hpp"
+#include "query_optimizer/rules/ReorderColumns.hpp"
 #include "query_optimizer/rules/StarSchemaHashJoinOrderOptimization.hpp"
 #include "query_optimizer/rules/SwapProbeBuild.hpp"
 #include "query_optimizer/strategy/Aggregate.hpp"
@@ -43,6 +44,10 @@
 
 namespace quickstep {
 namespace optimizer {
+
+DEFINE_bool(reorder_columns, true,
+            "Adjust the ordering of intermediate relations' columns to improve "
+            "copy performance.");
 
 DEFINE_bool(reorder_hash_joins, true,
             "If true, apply hash join order optimization to each group of hash "
@@ -108,6 +113,13 @@ P::PhysicalPtr PhysicalGenerator::optimizePlan() {
     rules.emplace_back(new PruneColumns());
   } else {
     rules.emplace_back(new SwapProbeBuild());
+  }
+  if (FLAGS_reorder_columns) {
+    // NOTE(jianqiao): This optimization relies on the fact that the intermediate
+    // relations all have SPLIT_ROW_STORE layouts. If this fact gets changed, the
+    // optimization algorithm may need to be updated and the performance impact
+    // should be re-evaluated.
+    rules.emplace_back(new ReorderColumns());
   }
   if (FLAGS_use_lip_filters) {
     rules.emplace_back(new AttachLIPFilters());
--- a/query_optimizer/rules/ReorderColumns.cpp
+++ b/query_optimizer/rules/ReorderColumns.cpp
@@ -1,0 +1,214 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ **/
+
+#include "query_optimizer/rules/ReorderColumns.hpp"
+
+#include <algorithm>
+#include <cstddef>
+#include <limits>
+#include <unordered_map>
+#include <vector>
+
+#include "query_optimizer/expressions/AttributeReference.hpp"
+#include "query_optimizer/expressions/ExprId.hpp"
+#include "query_optimizer/expressions/NamedExpression.hpp"
+#include "query_optimizer/physical/HashJoin.hpp"
+#include "query_optimizer/physical/PatternMatcher.hpp"
+#include "query_optimizer/physical/Physical.hpp"
+#include "query_optimizer/physical/PhysicalType.hpp"
+#include "query_optimizer/physical/Selection.hpp"
+#include "query_optimizer/physical/TableReference.hpp"
+
+#include "glog/logging.h"
+
+namespace quickstep {
+namespace optimizer {
+
+namespace E = ::quickstep::optimizer::expressions;
+namespace P = ::quickstep::optimizer::physical;
+
+P::PhysicalPtr ReorderColumns::apply(const P::PhysicalPtr &input) {
+  DCHECK(input->getPhysicalType() == P::PhysicalType::kTopLevelPlan);
+
+  return applyInternal(input, true);
+}
+
+P::PhysicalPtr ReorderColumns::applyInternal(const P::PhysicalPtr &input,
+                                             const bool lock_ordering) {
+  // We have to guarantee that the top level ordering of the columns remain
+  // unchanged so that the output columns are ordered as specified by the user.
+  // So here we use the flag "lock_ordering" to skip the first transformable
+  // node (i.e. the first Selection or HashJoin).
+  const bool is_not_transformable = !IsTransformable(input);
+  const bool skip_transform = lock_ordering || is_not_transformable;
+
+  if (skip_transform) {
+    std::vector<P::PhysicalPtr> new_children;
+    for (const P::PhysicalPtr &child : input->children()) {
+      new_children.emplace_back(applyInternal(child, lock_ordering && is_not_transformable));
+    }
+
+    if (new_children != input->children()) {
+      return input->copyWithNewChildren(new_children);
+    } else {
+      return input;
+    }
+  }
+
+  // Collect the maximal chain of transformable nodes.
+  std::vector<P::PhysicalPtr> nodes;
+  for (P::PhysicalPtr node = input; IsTransformable(node); node = node->children().front()) {
+    nodes.emplace_back(node);
+  }
+  // Arrange the nodes with bottom-up order.
+  std::reverse(nodes.begin(), nodes.end());
+
+  // A greedy algorithm that reorders the output attributes based on the GEN/KILL
+  // intervals. This algorithm works well with SSB/TPCH queries and is not likely
+  // to make the plans worse for whatever queries.
+  //
+  // Here is a brief explanation of the three data structure base/gen/kill.
+  //   (1) base: maps each attribute's id to its position in the BASE relation's
+  //             output attributes. Note that the base relation is the child
+  //             relation of nodes[0].
+  //   (2) gen:  maps each attribute's id to the MINIMUM index i such that the
+  //             attribute is among nodes[i]'s output attributes. I.e. node i
+  //             GENERATEs the attribute.
+  //   (3) kill: maps each attribute's id to the MAXIMUM index i such that the
+  //             attribute is among nodes[i]'s output attributes. I.e. node i+1
+  //             KILLs the attribute.
+  std::unordered_map<E::ExprId, std::size_t> base, gen, kill;
+
+  const P::PhysicalPtr base_node =
+      applyInternal(nodes.front()->children().front(), false);
+  const std::vector<E::AttributeReferencePtr> base_attrs =
+      base_node->getOutputAttributes();
+  for (std::size_t i = 0; i < base_attrs.size(); ++i) {
+    base.emplace(base_attrs[i]->id(), i);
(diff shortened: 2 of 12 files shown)


query_optimizer/rules/ReorderColumns.cpp
```diff
@@ -0,0 +1,205 @@
...
+      case P::PhysicalType::kSelection: {
+        project_expressions =
+            std::static_pointer_cast<const P::Selection>(node)->project_expressions();
+        break;
+      }
+      default:
+        LOG(FATAL) << "Unsupported physical type";
+    }
+
+    std::sort(project_expressions.begin(), project_expressions.end(), comparator);
```
zuyu: [Code style](https://google.github.io/styleguide/cppguide.html#Lambda_expressions): move the `comparator` lambda expression as a part of calling `sort`:

```
std::sort(v.begin(), v.end(), [](int x, int y) {
  return Weight(x) < Weight(y);
});
```

query_optimizer/PhysicalGenerator.cpp
```diff
@@ -109,6 +114,13 @@ P::PhysicalPtr PhysicalGenerator::optimizePlan() {
   } else {
     rules.emplace_back(new SwapProbeBuild());
   }
+  if (FLAGS_reorder_columns) {
+    // NOTE(jianqiao): This optimization relies on the fact that the intermediate
+    // relations all have SPLIT_ROW_STORE layouts. If this fact gets changed, the
+    // optimization algorithm may need to be updated and the performance impact
+    // should be re-evaluated.
```
zuyu: I suggest to check this assumption in `ReorderColumns::IsTransformable`, so we could remove this comment here.

query_optimizer/rules/ReorderColumns.cpp
```diff
@@ -0,0 +1,205 @@
...
+  // A greedy algorithm that reorders the output attributes based on the GEN/KILL
+  // intervals. This algorithm works well with SSB/TPCH queries and is not likely
+  // to make the plans worse for whatever queries.
+  std::unordered_map<E::ExprId, std::size_t> base, gen, kill;
+
+  const P::PhysicalPtr base_node =
+      applyInternal(nodes.front()->children().front(), false);
+  P::TableReferencePtr base_table;
+  if (P::SomeTableReference::MatchesWithConditionalCast(base_node, &base_table)) {
+  }
```
zuyu: Unused `base_table ` and do nothing in the if-case?

query_optimizer/rules/ReorderColumns.cpp
```diff
@@ -0,0 +1,205 @@
...
+  std::vector<P::PhysicalPtr> nodes;
+  for (P::PhysicalPtr node = input; IsTransformable(node); node = node->children().front()) {
+    nodes.emplace_back(node);
+  }
+  std::reverse(nodes.begin(), nodes.end());
+
+  // A greedy algorithm that reorders the output attributes based on the GEN/KILL
+  // intervals. This algorithm works well with SSB/TPCH queries and is not likely
+  // to make the plans worse for whatever queries.
+  std::unordered_map<E::ExprId, std::size_t> base, gen, kill;
```
zuyu: Please add more comments about the usage for both `gen` and `kill`, i.e., why calling `kill`?

query_optimizer/rules/ReorderColumns.cpp
```diff
@@ -0,0 +1,205 @@
...
+    const std::size_t rhs_kill = kill.at(rhs_id);
+    if (lhs_kill != rhs_kill) {
+      return lhs_kill < rhs_kill;
+    }
+
+    // Finally by the ordering in the base relaton.
+    const auto lhs_base_it = base.find(lhs_id);
+    const auto rhs_base_it = base.find(rhs_id);
+    const std::size_t lhs_base =
+        lhs_base_it == base.end() ? std::numeric_limits<std::size_t>::max()
```
zuyu: Optionally, if `lhs_base_it` is not found, we could simply return the size of `base`, instead of the max value.

relational_operators/HashJoinOperator.cpp
```diff
@@ -63,6 +65,9 @@ namespace quickstep {
 
 namespace {
 
+typedef std::vector<std::pair<tuple_id, tuple_id>> VectorOfPairs;
+typedef std::pair<std::vector<tuple_id>, std::vector<tuple_id>> PairOfVectors;
```
zuyu: Suggest to rename to `VectorOfTupleIdPair` and `PairOfTupleIdVector`, respectively.

relational_operators/HashJoinOperator.cpp
```diff
@@ -484,7 +565,38 @@ void HashInnerJoinWorkOrder::execute() {
   const relation_id build_relation_id = build_relation_.getID();
   const relation_id probe_relation_id = probe_relation_.getID();
 
-  for (std::pair<const block_id, std::pair<std::vector<tuple_id>, std::vector<tuple_id>>>
+  // Create a map of ValueAccessors and what attributes we want to pick from them.
+  std::vector<std::pair<ValueAccessor *, std::vector<attribute_id>>> accessor_attribute_map;
+  const std::size_t build_index = 0, probe_index = 1, temp_index = 2;
+  for (std::size_t i = 0; i < 3; ++i) {
```
zuyu: Please replace the magic number `3` with some constant like `kNumIndexes`.

Also please rename above three `indexes` to `kBuildIndex`, `kProbeIndex`, and `kTempIndex`, respectively.

relational_operators/HashJoinOperator.cpp
```diff
@@ -484,7 +565,38 @@ void HashInnerJoinWorkOrder::execute() {
...
+    accessor_attribute_map.emplace_back(
+        nullptr /* place holder */,
```
zuyu: Change the comment to `place holder for a late binding ValueAccessor`.

relational_operators/HashJoinOperator.cpp
```diff
@@ -504,123 +616,61 @@ void HashInnerJoinWorkOrder::execute() {
...
 
     // We also need a temp value accessor to store results of any scalar expressions.
     ColumnVectorsValueAccessor temp_result;
+    if (!non_trivial_expressions.empty()) {
+      // The getAllValuesForJoin function below needs joined tuple IDs as a
+      // vector of pair of (build-tuple-ID, probe-tuple-ID), and we have a pair
+      // of (build-tuple-IDs-vector, probe-tuple-IDs-vector). So we'll have to
+      // zip our two vectors together.
+      VectorOfPairs zipped_joined_tuple_ids;
+      zipped_joined_tuple_ids.reserve(build_tids.size());
```
zuyu: Please remove this line, as its effect is duplicated after executing the following `looped emplace_back`.

query_optimizer/rules/ReorderColumns.cpp
```diff
@@ -0,0 +1,205 @@
...
+  // So here we use the flag "lock_ordering" to skip the first transformable
+  // node (i.e. the first Selection or HashJoin).
+  const bool is_not_transformable = !IsTransformable(input);
+  const bool skip_transform = lock_ordering || is_not_transformable;
+  lock_ordering = lock_ordering && is_not_transformable;
+
+  if (skip_transform) {
+    std::vector<P::PhysicalPtr> new_children;
+    for (const P::PhysicalPtr &child : input->children()) {
+      new_children.emplace_back(applyInternal(child, lock_ordering));
```
zuyu: Implementation tricks: I think we could still mark the `lock_ordering` argument `const`, while keeping the same semantics by `new_children.emplace_back(applyInternal(child, lock_ordering && is_not_transformable));`.

query_optimizer/rules/ReorderColumns.cpp
```diff
@@ -0,0 +1,205 @@
...
+      return input;
+    }
+  }
+
+  // Collect the maximal chain of transformable nodes.
+  std::vector<P::PhysicalPtr> nodes;
+  for (P::PhysicalPtr node = input; IsTransformable(node); node = node->children().front()) {
+    nodes.emplace_back(node);
+  }
+  std::reverse(nodes.begin(), nodes.end());
```
zuyu: Please add a comment regarding the order in `nodes`, i.e., `bottom-up`.

query_optimizer/rules/ReorderColumns.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+  DISALLOW_COPY_AND_ASSIGN(ReorderColumns);
+};
+
+/** @} */
+
+}  // namespace optimizer
+}  // namespace quickstep
+
+#endif /* QUICKSTEP_QUERY_OPTIMIZER_RULES_REORDER_COLUMNS_HPP_ */
```
zuyu: Code style: `#endif  // QUICKSTEP_QUERY_OPTIMIZER_RULES_REORDER_COLUMNS_HPP_`

relational_operators/HashJoinOperator.cpp
```diff
@@ -20,6 +20,7 @@
 #include "relational_operators/HashJoinOperator.hpp"
 
 #include <algorithm>
+#include <map>
```
zuyu: Could you please point out the location of using `map` in this file? I could not find it.

relational_operators/HashJoinOperator.cpp
```diff
@@ -504,123 +616,61 @@ void HashInnerJoinWorkOrder::execute() {
...
+                                                          build_accessor.get(),
+                                                          probe_relation_id,
+                                                          probe_accessor,
+                                                          zipped_joined_tuple_ids));
       }
-      ++dest_attr;
     }
 
+    // We now create ordered value accessors for both build and probe side,
+    // using the joined tuple TIDs.
```
zuyu: Change from `TIDs` to `IDs`? I guess `TID` means `tuple id`.

storage/ValueAccessor.hpp
```diff
@@ -305,6 +305,21 @@ class ValueAccessor {
       const TupleIdSequence &id_sequence) = 0;
 
   /**
+   * @brief Create a new OrderedTupleIdSequenceAdapterValueAccessor that wraps
+   *        this ValueAccessor.
```
zuyu: Please add more comments explaining the `SharedOrderedTupleIdSequence`. Thanks!

query_optimizer/rules/ReorderColumns.cpp
jianqiao: Updated.

query_optimizer/rules/ReorderColumns.cpp
jianqiao: Updated.

query_optimizer/rules/ReorderColumns.cpp
jianqiao: Updated.

query_optimizer/rules/ReorderColumns.cpp
jianqiao: Removed.

query_optimizer/rules/ReorderColumns.cpp
jianqiao: That's true. I think it's ok to stick to max.

relational_operators/HashJoinOperator.cpp
jianqiao: Removed.

relational_operators/HashJoinOperator.cpp
jianqiao: Updated.

relational_operators/HashJoinOperator.cpp
jianqiao: `reserve` is for pre-allocating memory and avoids buffer reallocation during the loop.

relational_operators/HashJoinOperator.cpp
jianqiao: Updated.

query_optimizer/rules/ReorderColumns.hpp
jianqiao: Updated.

storage/ValueAccessor.hpp
jianqiao: `SharedOrderedTupleIdSequence` has its header comments at `ValueAccessor.hpp` line 548.

jianqiao: `OrderedTupleIdSequenceAdapterValueAccessor` has its header comments at `ValueAccessor.hpp` line 548.
`OrderedTupleIdSequence` has its parameter comment below.

relational_operators/HashJoinOperator.cpp
jianqiao: Updated.

relational_operators/HashJoinOperator.cpp
jianqiao: Updated.

query_optimizer/PhysicalGenerator.cpp
jianqiao: Currently this cannot be checked because the temporary relations' layouts are generated in `ExecutionGenerator` when they get created.

In the future we will reason about the layouts during query optimization and add corresponding data structures.

query_optimizer/rules/ReorderColumns.cpp
jianqiao: Added a TODO there.

storage/ValueAccessor.hpp
zuyu: I understand `OrderedTupleIdSequence`, but was wondering what does `Shared` mean here.

PR #174, opened 2017-01-30
Push down disjunctive predicates to filter small-cardinality stored relation early.

jianqiao:
This PR implements an optimization (physical plan transformation) that pushes down disjunctive predicate to filter stored relations early when proper conditions are met.

Here we elaborate the conditions. Let
```
P = p_{1,1} AND ... AND p_{1, m_1}
    OR
    ...
    OR
    p_{n,1} AND ... AND p_{n, m_n}
```
be a predicate in _disjunctive normal form_.
 
Now consider each small-cardinality relation R, if for each `i` in `1..n`, there exists at least one predicate `p_{i, k_i}` that is applicable to R. Then we can construct a new predicate
```
P' = p_{1, k_1} OR ... OR p_{n, k_n}
```
and push down `P'` to be applied to R.
 
Also, if any conjunctive component in `P` contains more than one predicate that is applicable to R, then we can combine all these applicable predicates as a conjunctive component in `P'`.
 
Finally, note that if there exists a conjunctive component that contains no predicate applicable to R. Then the condition fails and we cannot do a push down for R.

This optimization improves the performance of TPC-H Q07 from ~17 seconds to ~4 seconds, with SF100 on a cloudlab machine.

### DIFF
--- a/query_optimizer/PhysicalGenerator.cpp
+++ b/query_optimizer/PhysicalGenerator.cpp
@@ -28,6 +28,7 @@
 #include "query_optimizer/physical/Physical.hpp"
 #include "query_optimizer/rules/AttachLIPFilters.hpp"
 #include "query_optimizer/rules/PruneColumns.hpp"
+#include "query_optimizer/rules/PushDownLowCostDisjunctivePredicate.hpp"
 #include "query_optimizer/rules/ReorderColumns.hpp"
 #include "query_optimizer/rules/StarSchemaHashJoinOrderOptimization.hpp"
 #include "query_optimizer/rules/SwapProbeBuild.hpp"
@@ -108,12 +109,22 @@ P::PhysicalPtr PhysicalGenerator::generateInitialPlan(
 P::PhysicalPtr PhysicalGenerator::optimizePlan() {
   std::vector<std::unique_ptr<Rule<P::Physical>>> rules;
   rules.emplace_back(new PruneColumns());
+
+  // TODO(jianqiao): It is possible for PushDownLowCostDisjunctivePredicate to
+  // generate two chaining Selection nodes that can actually be fused into one.
+  // Note that currently it is okay to have the two Selections because they are
+  // applied to a small cardinality stored relation, which is very light-weight.
+  // However it is better to have a FuseSelection optimization (or even a more
+  // general FusePhysical optimization) in the future.
+  rules.emplace_back(new PushDownLowCostDisjunctivePredicate());
+
   if (FLAGS_reorder_hash_joins) {
     rules.emplace_back(new StarSchemaHashJoinOrderOptimization());
     rules.emplace_back(new PruneColumns());
   } else {
     rules.emplace_back(new SwapProbeBuild());
   }
+
   if (FLAGS_reorder_columns) {
     // NOTE(jianqiao): This optimization relies on the fact that the intermediate
     // relations all have SPLIT_ROW_STORE layouts. If this fact gets changed, the
@@ -121,6 +132,10 @@ P::PhysicalPtr PhysicalGenerator::optimizePlan() {
     // should be re-evaluated.
     rules.emplace_back(new ReorderColumns());
   }
+
+  // NOTE(jianqiao): Adding rules after AttachLIPFilters requires extra handling
+  // of LIPFilterConfiguration for transformed nodes. So currently it is suggested
+  // that all the new rules be placed before this point.
   if (FLAGS_use_lip_filters) {
     rules.emplace_back(new AttachLIPFilters());
   }
--- a/query_optimizer/rules/PushDownLowCostDisjunctivePredicate.cpp
+++ b/query_optimizer/rules/PushDownLowCostDisjunctivePredicate.cpp
@@ -1,0 +1,225 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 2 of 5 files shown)


query_optimizer/PhysicalGenerator.cpp
```diff
@@ -27,6 +27,7 @@
 #include "query_optimizer/logical/Logical.hpp"
 #include "query_optimizer/physical/Physical.hpp"
 #include "query_optimizer/rules/AttachLIPFilters.hpp"
+#include "query_optimizer/rules/PushDownLowCostDisjunctivePredicate.hpp"
```
zuyu: Code style: sort in alphabetic order, so move this line after the next.

query_optimizer/PhysicalGenerator.cpp
```diff
@@ -108,6 +109,7 @@ P::PhysicalPtr PhysicalGenerator::generateInitialPlan(
 P::PhysicalPtr PhysicalGenerator::optimizePlan() {
   std::vector<std::unique_ptr<Rule<P::Physical>>> rules;
   rules.emplace_back(new PruneColumns());
+  rules.emplace_back(new PushDownLowCostDisjunctivePredicate());
```
zuyu: Could we have some comments regarding the order of `rules`, i.e., why put `PushDownLowCostDisjunctivePredicate` here?

If I understand correctly, `PushDownLowCostDisjunctivePredicate` may generate new `Physical Select`s. Would the rest of `rules` eliminate multiple `Physical Select`s into one?

query_optimizer/rules/PushDownLowCostDisjunctivePredicate.cpp
```diff
@@ -0,0 +1,218 @@
...
+    return input;
+  }
+}
+
+void PushDownLowCostDisjunctivePredicate::collectApplicablePredicates(
+    const physical::PhysicalPtr &input) {
+  P::TableReferencePtr table_reference;
+  if (P::SomeTableReference::MatchesWithConditionalCast(input, &table_reference)) {
+    // Consider only stored relations with small cardinality as targets.
+    if (cost_model_->estimateCardinality(input) <= 100u) {
```
zuyu: Just curious about `100u` here. My concern is that the result of `estimateCardinality` for a small relation is two orders bigger than the fact; I have seen it in the `query_optimizer execution_generator` unit test.

Also, I'd suggest to use a `gflags` for `100u` to replace `kCardinalityThreshold` defined in the header file.

query_optimizer/rules/PushDownLowCostDisjunctivePredicate.cpp
```diff
@@ -0,0 +1,218 @@
...
+    const std::vector<E::PredicatePtr> predicates) {
+  DCHECK_GE(predicates.size(), 1u);
+  if (predicates.size() == 1) {
+    return predicates.front();
+  } else {
+    return E::LogicalOr::Create(predicates);
+  }
+}
+
+
```
zuyu: Remove this extra empty line.

query_optimizer/rules/PushDownLowCostDisjunctivePredicate.hpp
```diff
@@ -0,0 +1,118 @@
...
+  struct PredicateInfo {
+    PredicateInfo() {}
+    inline void add(expressions::PredicatePtr predicate) {
+      predicates.emplace_back(predicate);
+    }
+    std::vector<expressions::PredicatePtr> predicates;
+  };
+
+  void collectApplicablePredicates(const physical::PhysicalPtr &input);
+  physical::PhysicalPtr attachPredicates(const physical::PhysicalPtr &input) const;
```
zuyu: Please add an empty line between two above methods.

query_optimizer/PhysicalGenerator.cpp
jianqiao: The three rules blocks from `PushDownLowCostDisjunctivePredicate` to `ReorderColumns` can be arranged in any order.

`PushDownLowCostDisjunctivePredicate` will not generate unnecessary `Selection`s. And currently in the physical generate there's no rule that fuses `Selection`s.

I add a comment to indicate that currently new rules should better be added before `AttachLIPFilters` because rules after that needs extra handling of `LIPFilterConfiguration` for transformed nodes.

query_optimizer/PhysicalGenerator.cpp
jianqiao: Updated.

query_optimizer/rules/PushDownLowCostDisjunctivePredicate.cpp
jianqiao: It is okay to be conservative and the stat is more accurate if the `\analyze` command gets executed.

Added a gflag.

query_optimizer/rules/PushDownLowCostDisjunctivePredicate.hpp
jianqiao: Updated.

query_optimizer/rules/PushDownLowCostDisjunctivePredicate.cpp
jianqiao: Updated.

query_optimizer/PhysicalGenerator.cpp
zuyu: I think [here](https://github.com/apache/incubator-quickstep/pull/174/files#diff-ca3b59cc48fbc383291bdfefaf5de128R188) we add a new `Physical::Select`.

jianqiao: Yes it will only create necessary `Selection`s. I.e. there won't be two selections in a chain that can be fused into one.

jianqiao: Oh I made a mistake that actually the `InjectFilterJoin` optimization will not generate unnecessary `Selection`s, but `PushDownLowCostDisjunctivePredicate` may generate two `Selection`s in a chain. Currently this is a not an issue since the `Selection`s for tables with cardinality around 100 are very light-weight. We may add a `FuseSelection` optimization later.

zuyu: Please add a `NOTE` comment regarding two `Selection`s after applying this rule.

jianqiao: Updated.

PR #177, opened 2017-02-02
Reduce the number of group-by attributes by pulling tables up aggregation.

jianqiao:
This PR implements an optimization (physical plan transformation) that pulls a table up an aggregation if many of that table's attributes serve as group-by attributes in the aggregation. We do the optimization because it is relatively slow to aggregate with a large set of group-by attributes, as well as to avoid copying the table's many attributes all the way up a chain of operators.

This optimization improves the performance of TPC-H Q10 to ~5.7s, with scale factor 100 on a cloudlab machine. Below shows the physical plans for TPC-H Q10 with/without the optimization.

[WithoutReduce.pdf](https://github.com/apache/incubator-quickstep/files/753410/WithoutReduce.pdf)

[WithReduce.pdf](https://github.com/apache/incubator-quickstep/files/753411/WithReduce.pdf)

### DIFF
--- a/query_optimizer/Optimizer.cpp
+++ b/query_optimizer/Optimizer.cpp
@@ -30,10 +30,11 @@ void Optimizer::generateQueryHandle(const ParseStatement &parse_statement,
                                     OptimizerContext *optimizer_context,
                                     QueryHandle *query_handle) {
   LogicalGenerator logical_generator(optimizer_context);
+  PhysicalGenerator physical_generator(optimizer_context);
   ExecutionGenerator execution_generator(catalog_database, query_handle);
 
   execution_generator.generatePlan(
-      physical_generator_.generatePlan(
+      physical_generator.generatePlan(
           logical_generator.generatePlan(*catalog_database, parse_statement)));
 }
 
--- a/query_optimizer/rules/ReduceGroupByAttributes.cpp
+++ b/query_optimizer/rules/ReduceGroupByAttributes.cpp
@@ -1,0 +1,217 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ **/
+
+#include "query_optimizer/rules/ReduceGroupByAttributes.hpp"
+
+#include <algorithm>
+#include <map>
+#include <vector>
+#include <unordered_set>
+#include <utility>
+
+#include "catalog/CatalogRelation.hpp"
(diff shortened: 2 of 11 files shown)


query_optimizer/rules/ReduceGroupByAttributes.cpp
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  P::AggregatePtr aggregate;
+  if (!P::SomeAggregate::MatchesWithConditionalCast(input, &aggregate) ||
+      aggregate->grouping_expressions().size() <= 1u) {
+    return input;
+  }
+
+  // Divide the group-by attributes into groups based on their source table.
+  std::map<P::TableReferencePtr, std::vector<E::AttributeReferencePtr>> table_attributes;
```
zuyu: Could we use a faster `unordered_map` instead?

jianqiao:
This optimization is somehow different from `AggregatePushDown`. I updated the example in this PR's description.

query_optimizer/Optimizer.cpp
```diff
@@ -30,10 +30,11 @@ void Optimizer::generateQueryHandle(const ParseStatement &parse_statement,
                                     OptimizerContext *optimizer_context,
                                     QueryHandle *query_handle) {
   LogicalGenerator logical_generator(optimizer_context);
+  PhysicalGenerator physical_generator(optimizer_context);
```
jianqiao: This might be done later together with the same change to `LogicalGenerator`.

query_optimizer/rules/ReduceGroupByAttributes.cpp
jianqiao: We would expect tables attributes to contain a small number of entries (# of storage tables for the attributes in one group-by clause, typically 1 to 3). Also consider that there an iteration of all the map elements below. It might be better to use a `std::map` here. Anyway it won't make even a tiny performance difference to use either data structure here.

pateljm:
@jianqiao Another neat optimization! I also like how your PRs are clean. Thank you for adding the pdf, which makes understanding this optimization super easy. 

@zuyu Great comments. I'd like to suggest that given the scope of the PR, the requests you make be rolled into a future PR.

I'm happy to close this PR -- so @jianqiao feel free to assign this to me.

query_optimizer/rules/ReduceGroupByAttributes.hpp
```diff
@@ -0,0 +1,143 @@
...
+      if (lhs->maximum_size != rhs->maximum_size) {
+        return lhs->maximum_size < rhs->maximum_size;
+      }
+      return lhs->attribute->id() < rhs->attribute->id();
+    }
+
+    const expressions::AttributeReferencePtr attribute;
+    const bool is_unique;
+    const bool is_fixed_length;
+    const std::size_t maximum_size;
```
jianqiao: `maximum_size` is the maximum byte size of the attribute's values. It is used for comparison -- i.e. we prefer attributes with smaller maximum-byte-sizes as group-by keys.

query_optimizer/rules/ReduceGroupByAttributes.hpp
```diff
@@ -0,0 +1,143 @@
...
+ *   HashJoin(
+ *     [probe relation]: Aggregate(
+ *                         [input relation]: S
+ *                         [aggregate expression]: SUM(v) AS sum_v
+ *                         [group-by attribute]: u
+ *                       ) AS T
+ *     [build relation]: R
+ *     [join expression]: T.u = R.x
+ *     [project attributes]: sum_v, x, y, z
+ *   )
```
jianqiao: I will update the header later.

query_optimizer/rules/ReduceGroupByAttributes.hpp
zuyu: Then, please rename to `maximum_byte_length`, as defined in `Type`.

PR #179, opened 2017-02-05
QUICKSTEP-70-71 Improve aggregation performance

jianqiao:
This PR implements two features that improve aggregation performance:
1. Adds `CollisionFreeVectorTable` to support specialized high-performance aggregation.
2. Adds support for aggregation copy elision that we only materialize intermediate results for non-trivial expressions.

For feature 1, when the group-by attribute is a range-bounded single attribute of `INT` or `LONG` type. We can use a vector of type `std::vector<std::atomic<StateT>>` to store the aggregation states, where `StateT` is the aggregation state type (currently restricted to `LONG` and `DOUBLE`). Then during aggregation, for each tuple, we locate the aggregation state with the group-by key's value as index to the state vector, and concurrently update the state with C++'s atomic primitives.

For feature 2, note that the current implementation of aggregation always creates a `ColumnVectorsValueAccessor` to store the results of ALL the input expressions. However, we can avoid the creation of a column vector (thus avoiding copying values into the column vector) if the aggregation is on a simple attribute, e.g. `SUM(x)`. Thus by PR, when performing aggregation we prepare two input `ValueAccessor`s: one BASE accessor that is created directly from the input relation's storage block, and one DERIVED accessor that is the temporary result `ColumnVectorsValueAccessor`. Each aggregation argument may be from the base accessor (meaning that it is a simple attribute) or from the derived accessor (meaning that it is a non-trivial expression that gets evaluated). The two accessors are then properly handled in aggregation handles and aggregation hash tables.

**Main changes:**
`expressions/aggregation`: Updated the aggregation handles to support copy elision. Also did some cleanups.
`relational_operators`: Added `InitializeAggregationOperator` to support parallel initialization of the aggregation state (just `memset` the memory to 0) -- because it takes a relatively long time to do the initialization with single thread if the aggregation hash table is large.
`storage`: Added `CollisionFreeVectorTable`. Renamed `FastHashTable` to `PackedPayloadHashTable`, made it support copy elision, and cleaned up the class to remove unused methods. Refactored `AggregationOperationState` to support copy elision and support the new aggregation. Moved aggregation code out of `StorageBlock`.

This PR significantly improves some TPC-H queries' performance. For example, it improves TPC-H Q18 from ~27.5s to ~3.5s, with scale factor 100 on a cloudlab machine.

Below shows the TPC-H performance (scale factor 100 on a cloudlab machine) with recently committed optimizations up to this point:

|  **TPCH SF100** | **master (ms)** | **w/ optimizations (ms)** |
|  ------ | ------ | ------ |
|  Q01 | 13629 | 11221 |
|  Q02 | 537 | 460 |
|  Q03 | 4824 | 4124 |
|  Q04 | 2185 | 2203 |
|  Q05 | 5517 | 5282 |
|  Q06 | 399 | 401 |
|  Q07 | 18563 | 3456 |
|  Q08 | 1746 | 899 |
|  Q09 | 7247 | 5586 |
|  Q10 | 6745 | 5665 |
|  Q11 | 1053 | 247 |
|  Q12 | 1713 | 1698 |
|  Q13 | 22896 | 15582 |
|  Q14 | 805 | 745 |
|  Q15 | 897 | 431 |
|  Q16 | 9942 | 9158 |
|  Q17 | 1588 | 1117 |
|  Q18 | 27459 | 3507 |
|  Q19 | 1711 | 1609 |
|  Q20 | 1204 | 1014 |
|  Q21 | 8671 | 7886 |
|  Q22 | 6178 | 724 |
|  **Total** | **145509** | **83016** |


zuyu:
Hi @jianqiao,

A quick question on Feature 1 using a vector-based aggregation: for a group-by w/ a known bounded range (i.e., the min and max value), do we always choose this approach over the hash-based, or depending on the range size (i.e., if the range is too wide, we may fall back to the hash-based)? Thanks!

jianqiao:
Currently there is a gflag for setting the range upbound:
https://github.com/apache/incubator-quickstep/pull/179/files#diff-3d4b5df01ed8edbe255e096eefbfc342R440

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -371,6 +378,109 @@ void ExecutionGenerator::dropAllTemporaryRelations() {
...
+    std::size_t *max_num_groups) const {
+#ifdef QUICKSTEP_DISTRIBUTED
+  // Currently we cannot do this fast path with the distributed setting. See
+  // the TODOs at InitializeAggregationOperator::getAllWorkOrderProtos() and
+  // FinalizeAggregationOperator::getAllWorkOrderProtos().
+  return false;
+#endif
+
+  // Supports only single group-by key.
+  if (aggregate->grouping_expressions().size() != 1) {
```
jianqiao: Yes we will have a followup PR to improve TPC-H Q01.

expressions/aggregation/AggregationConcreteHandle.hpp
```diff
@@ -61,7 +61,7 @@ class HashTableStateUpserterFast {
    *        table. The corresponding state (for the same key) in the destination
    *        hash table will be upserted.
    **/
-  HashTableStateUpserterFast(const HandleT &handle,
+  HashTableStateUpserter(const HandleT &handle,
                              const std::uint8_t *source_state)
```
jianqiao: Updated.

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -371,6 +378,109 @@ void ExecutionGenerator::dropAllTemporaryRelations() {
...
+      max_cpp_value = max_value.getLiteral<int>();
+      break;
+    }
+    case TypeID::kLong: {
+      min_cpp_value = min_value.getLiteral<std::int64_t>();
+      max_cpp_value = max_value.getLiteral<std::int64_t>();
+      break;
+    }
+    default:
+      return false;
```
jianqiao: We can support more types later. For any type/any number of group-by keys, if we can define a one-to-one mapping function that maps the keys to range-bounded integers, then this aggregation is applicable.

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -371,6 +378,109 @@ void ExecutionGenerator::dropAllTemporaryRelations() {
...
+  for (const auto &agg_expr : aggregate->aggregate_expressions()) {
+    const E::AggregateFunctionPtr agg_func =
+        std::static_pointer_cast<const E::AggregateFunction>(agg_expr->expression());
+
+    if (agg_func->is_distinct()) {
+      return false;
+    }
+
+    // TODO(jianqiao): Support AggregationID::AVG.
+    switch (agg_func->getAggregate().getAggregationID()) {
```
jianqiao: Updated.

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -371,6 +378,109 @@ void ExecutionGenerator::dropAllTemporaryRelations() {
...
+        return false;
+    }
+
+    const auto &arguments = agg_func->getArguments();
+    if (arguments.size() > 1) {
+      return false;
+    }
+
+    if (arguments.size() == 1) {
+      switch (arguments.front()->getValueType().getTypeID()) {
```
jianqiao: Updated.

query_optimizer/ExecutionGenerator.hpp
```diff
@@ -427,7 +445,7 @@ class ExecutionGenerator {
   /**
    * @brief The cost model to use for estimating aggregation hash table size.
    */
-  std::unique_ptr<cost::CostModel> cost_model_for_aggregation_;
+  std::unique_ptr<cost::StarSchemaSimpleCostModel> cost_model_for_aggregation_;
```
jianqiao: We need some extra analysis done in `StarSchemaSimpleCostModel`. I will refactor the analysis part out of `StarSchemaSimpleCostModel` into `AnalysisUtil` in a follow-up PR.

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -1495,9 +1607,28 @@ void ExecutionGenerator::convertAggregate(
...
+    const std::size_t estimated_num_groups =
+        cost_model_for_aggregation_->estimateNumGroupsForAggregate(physical_plan);
+
+    std::size_t max_num_groups;
+    const bool can_use_collision_free_aggregation =
+        canUseCollisionFreeAggregation(physical_plan,
+                                       estimated_num_groups,
+                                       &max_num_groups);
+
+    if (can_use_collision_free_aggregation) {
```
jianqiao: Updated.

storage/PackedPayloadHashTable.cpp
```diff
@@ -0,0 +1,463 @@
...
+  // Pick out a prime number of slots and calculate storage requirements.
+  std::size_t num_slots_tmp =
+      get_next_prime_number(num_entries * kHashTableLoadFactor);
+  std::size_t required_memory =
+      sizeof(Header) + num_slots_tmp * sizeof(std::atomic<std::size_t>) +
+      (num_slots_tmp / kHashTableLoadFactor) *
+          (bucket_size_ + key_manager_.getEstimatedVariableKeySize());
+  std::size_t num_storage_slots =
+      this->storage_manager_->SlotsNeededForBytes(required_memory);
+  if (num_storage_slots == 0) {
```
jianqiao: All the code in `PackedPayloadHashTable` is from the original `HashTable` implementation. We may have an overall code-style revision later.

storage/AggregationOperationState.hpp
```diff
@@ -156,6 +152,29 @@ class AggregationOperationState {
...
+   **/
+  std::size_t getNumInitializationPartitions() const;
+
+  /**
+   * @brief Get the number of partitions to be used for finalizing the
+   *        aggregation.
+   *
+   * @return The number of partitions to be used for finalizing the aggregation.
+   **/
+  std::size_t getNumFinalizationPartitions() const;
```
jianqiao: The number of partitions can be different for initialization and finalization.

jianqiao: Currently initialization is for `memset` the storage memory. It is much faster than finalization so we expect to have a smaller number of initialization partitions than finalization partitions if the storage memory is not large.

E.g. suppose the vector table has 1 million entries of `int` keys and `std::int64_t` states -- amount to 12MB storage memory. Then we may want 80 partitions to finalize the table, but would not create 80 work orders to initialize the 12MB memory.

zuyu: I get your point, but for the distributed version, my concern is that we have to schedule the finalize work order with a partition to the same node that executes the initialize work order. Thus, your approach makes it harder to implement in the distributed version.

On the other hand, we could measure the overheads of the initialize work order, and I guess it is related small, compared to the whole aggregation processing time.

storage/AggregationOperationState.cpp
```diff
@@ -353,187 +353,286 @@ bool AggregationOperationState::ProtoIsValid(
...
-void AggregationOperationState::aggregateBlockSingleState(
-    const block_id input_block) {
-  // Aggregate per-block state for each aggregate.
-  std::vector<std::unique_ptr<AggregationState>> local_state;
+void AggregationOperationState::initialize(const std::size_t partition_id) {
+  if (is_aggregate_collision_free_) {
+    static_cast<CollisionFreeVectorTable *>(
+        collision_free_hashtable_.get())->initialize(partition_id);
+  } else {
+    LOG(FATAL) << "AggregationOperationState::initializeState() "
```
jianqiao: If the code reaches here, then the system will probably actually crash. I.e. it is like a `CHECK` but we still put the `else` branch here for further updates.

storage/AggregationOperationState.cpp
```diff
@@ -353,187 +353,286 @@ bool AggregationOperationState::ProtoIsValid(
...
-    InsertDestination *output_destination) {
-  if (group_by_list_.empty()) {
-    finalizeSingleState(output_destination);
+std::size_t AggregationOperationState::getNumInitializationPartitions() const {
+  if (is_aggregate_collision_free_) {
+    return static_cast<CollisionFreeVectorTable *>(
+        collision_free_hashtable_.get())->getNumInitializationPartitions();
   } else {
-    finalizeHashTable(output_destination);
+    return 0u;
```
jianqiao: Currently we create the `InitializeAggregationOperator` only for collision-free aggregations.

storage/AggregationOperationState.cpp
```diff
@@ -353,187 +353,286 @@ bool AggregationOperationState::ProtoIsValid(
...
+  } else  {
+    return 1u;
   }
 }
 
-void AggregationOperationState::aggregateBlockSingleState(
-    const block_id input_block) {
-  // Aggregate per-block state for each aggregate.
-  std::vector<std::unique_ptr<AggregationState>> local_state;
+void AggregationOperationState::initialize(const std::size_t partition_id) {
```
jianqiao: The "partition" here is somehow not related to catalog or relations. I.e. it refers to the manual segmentation of the memory / hash table entries but not the partitioning of storage blocks.

storage/AggregationOperationState.cpp
```diff
@@ -80,148 +83,145 @@ AggregationOperationState::AggregationOperationState(
...
+      // TODO(jianqiao): estimated_num_entries is quite inaccurate for
+      // estimating the number of entries in the distinctify hash table.
+      // We need to estimate for each distinct aggregation an
+      // estimated_num_distinct_keys value during query optimization.
+      distinctify_hashtables_.emplace_back(
+          AggregationStateHashTableFactory::CreateResizable(
+              *distinctify_hash_table_impl_types_it,
+              key_types,
+              estimated_num_entries,
+              {},
```
jianqiao: Updated.

storage/AggregationOperationState.hpp
jianqiao: In the distributed scenario, you may need the table to be "really partitioned" (i.e. divided into several tables) in a different sense of what partitions refer to here. The partitions here cannot be scheduled to difference nodes since they have to update the same table.

So later when we make `CollisionFreeVectorTable` support really partitioned input relations, we will can rename all the "partitions" here to some other name, say `segments`.

jianqiao:
For the question about `PartitionedHashTablePool` and `HashTablePool`. Note that their use patterns are different so perhaps it is not natural to merge them into one class.

`PartitionedHashTablePool` creates **a fixed number** of hash tables **on its construction**. The use pattern is that every `AggregationWorkOrder` updates **all** of these hash tables and every `FinalizeAggregationWorkOrder` updates one of these hash tables.

`HashTablePool` creates hash tables **on demand**. The current use pattern is that every `AggregationWorkOrder` checkouts **exclusively** one hash table, updates the hash table, and returns the hash table back to the pool. Then only one `FinalizeAggregationWorkOrder` is created to merge all the tables in the pool to the final hash table.

relational_operators/InitializeAggregationOperator.cpp
```diff
@@ -0,0 +1,72 @@
...
+          op_index_);
+    }
+    started_ = true;
+  }
+  return true;
+}
+
+// TODO(quickstep-team) : Think about how the number of partitions could be
+// accessed in this function. Until then, we can't use partitioned aggregation
+// initialization with the distributed version.
```
jianqiao: The "partition" here is kind of inner-table partitioning that is not related to the partitioning of storage blocks. To get the "number of partitions" here, we need the "AggregationOperationState" object, which is not available at this moment.

One solution is to calculate the # of partitions value before creating the `AggregationOperationState` object. That requires quite some refactoring. We can see if we need to do that later.

zuyu: So, here we don't use one or multiple catalog attribute as the partition key in the inner-table? Then, how do we do partition? Thanks.

storage/AggregationOperationState.cpp
zuyu: May two segmented hash tables have the same hashed key?

relational_operators/InitializeAggregationOperator.cpp
jianqiao: Currently no. For initialization, say we have some the allocated table memory pointer as `m` with size `40MB`, then we just create 10 work orders, so that work order `i` do `memset(m + 4M * i, 0, 4M)`.

storage/AggregationOperationState.cpp
jianqiao: No. Every segment contains an exclusive subset of buckets. Every bucket corresponds to a unique key.

storage/AggregationOperationState.cpp
```diff
@@ -353,187 +353,286 @@ bool AggregationOperationState::ProtoIsValid(
...
-  }
-
-  // This holds values of all the GROUP BY attributes so that the can be reused
-  // across multiple aggregates (i.e. we only pay the cost of evaluatin the
-  // GROUP BY expressions once).
-  std::vector<std::unique_ptr<ColumnVector>> reuse_group_by_vectors;
-
+void AggregationOperationState::mergeSingleState(
+    const std::vector<std::unique_ptr<AggregationState>> &local_state) {
+  DEBUG_ASSERT(local_state.size() == single_states_.size());
```
jianqiao: Updated.

zuyu:
Please resync with the master branch, and I will merge it. Thanks.

relational_operators/WorkOrderFactory.cpp
```diff
@@ -186,6 +186,7 @@ WorkOrder* WorkOrderFactory::ReconstructFromProto(const serialization::WorkOrder
       LOG(INFO) << "Creating FinalizeAggregationWorkOrder in Shiftboss " << shiftboss_index;
       return new FinalizeAggregationWorkOrder(
           proto.query_id(),
+          0uL,
```
jianqiao: Added a TODO comment:

```
// TODO(quickstep-team): Handle inner-table partitioning in the distributed setting.
```

storage/AggregationOperationState.cpp
```diff
@@ -556,80 +655,83 @@ void AggregationOperationState::finalizeSingleState(
...
+      new NativeColumnVector(*key_type, max_length));
+  hash_table->finalizeKey(partition_id, key_cv.get());
+  complete_result.addColumn(key_cv.release());
+
+  for (std::size_t i = 0; i < handles_.size(); ++i) {
+    const Type *result_type = handles_[i]->getResultType();
+    DCHECK(NativeColumnVector::UsableForType(*result_type));
+
+    std::unique_ptr<NativeColumnVector> result_cv(
+        new NativeColumnVector(*result_type, max_length));
```
jianqiao: Updated.

jianqiao:
Just rebased on master.

jianqiao:
There is a `CMakeLists` to be updated -- do not merge at this moment.

jianqiao:
Updated, and tested locally.

pateljm:
Very impressive algorithmic change @jianqiao !!

PR #180, opened 2017-02-05
Fix the bug with SingleIdentityHashFilter when filter_cardinality is 0.

jianqiao:
This PR fixes the problem with `SingleIdentityHashFilter` when the estimated filter cardinality is 0 (by setting the filter cardinality to at least 64).

### DIFF
--- a/query_optimizer/rules/AttachLIPFilters.cpp
+++ b/query_optimizer/rules/AttachLIPFilters.cpp
@@ -19,6 +19,7 @@
 
 #include "query_optimizer/rules/AttachLIPFilters.hpp"
 
+#include <algorithm>
 #include <map>
 #include <set>
 #include <unordered_set>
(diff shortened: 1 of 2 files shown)


jianqiao:
Assigned to @zuyu.

zuyu:
Alternatively, how about not adding a filter when the estimated cardinality is zero?

jianqiao:
Estimated cardinality being zero means that we expect the relation to be empty or have a small number of tuples. In either case, the created filter is effective and can be used to pre-filter other relations. Otherwise if we don't create such a filter, then there would be no optimization at all.

query_optimizer/rules/AttachLIPFilters.cpp
```diff
@@ -128,7 +129,7 @@ void AttachLIPFilters::attachLIPFilters(
           lip_filter_configuration_->addBuildInfo(
               P::SingleIdentityHashFilterBuildInfo::Create(
                   pair.second->source_attribute,
-                  pair.second->estimated_cardinality * 8),
+                  std::max(64uL, pair.second->estimated_cardinality * 8u)),
```
zuyu: Could we add a flag here instead of a hard-coded value?

jianqiao: It seems that we won't need to change this value. If there are scenarios later that the configuration of this value makes a difference, then we can add a flag for that.

PR #181, opened 2017-02-05
Added limited optimizer support for Partitioned Hash Joins.

zuyu:
This PR allows repartition for non-stored, partitioned input relations for hash joins.

Repartition will generate the best partition attribute if both input relations have partitions that do not meet the join condition. Otherwise, repartition on one input relation will reuse the partition attribute from the other that meets the join condition.

### DIFF
--- a/query_optimizer/ExecutionGenerator.cpp
+++ b/query_optimizer/ExecutionGenerator.cpp
@@ -138,8 +138,12 @@
 #include "gflags/gflags.h"
 #include "glog/logging.h"
 
+using std::find;
+using std::make_unique;
 using std::move;
+using std::size_t;
 using std::static_pointer_cast;
+using std::swap;
 using std::unique_ptr;
 using std::unordered_map;
 using std::vector;
@@ -162,6 +166,8 @@ static const volatile bool aggregate_hashtable_type_dummy
                                     &ValidateHashTableImplTypeString);
 
 DEFINE_bool(parallelize_load, true, "Parallelize loading data files.");
+
+DEFINE_uint64(num_repartitions, 4, "Number of repartitions for a hash join.");
 
 namespace E = ::quickstep::optimizer::expressions;
 namespace P = ::quickstep::optimizer::physical;
@@ -428,7 +434,8 @@ void ExecutionGenerator::convertTableReference(
       std::piecewise_construct,
       std::forward_as_tuple(physical_table_reference),
       std::forward_as_tuple(CatalogRelationInfo::kInvalidOperatorIndex,
-                            catalog_relation));
+                            catalog_relation,
+                            QueryContext::kInvalidInsertDestinationId));
 }
 
 void ExecutionGenerator::convertSample(const P::SamplePtr &physical_sample) {
@@ -468,8 +475,9 @@ void ExecutionGenerator::convertSample(const P::SamplePtr &physical_sample) {
       std::piecewise_construct,
       std::forward_as_tuple(physical_sample),
       std::forward_as_tuple(sample_index,
-                            output_relation));
-  temporary_relation_info_vec_.emplace_back(sample_index, output_relation);
+                            output_relation,
+                            insert_destination_index));
+  temporary_relation_info_vec_.emplace_back(sample_index, output_relation, insert_destination_index);
 }
 
 bool ExecutionGenerator::convertSimpleProjection(
@@ -602,8 +610,9 @@ void ExecutionGenerator::convertSelection(
       std::piecewise_construct,
       std::forward_as_tuple(physical_selection),
       std::forward_as_tuple(select_index,
-                            output_relation));
-  temporary_relation_info_vec_.emplace_back(select_index, output_relation);
+                            output_relation,
+                            insert_destination_index));
+  temporary_relation_info_vec_.emplace_back(select_index, output_relation, insert_destination_index);
 
   if (lip_filter_generator_ != nullptr) {
     lip_filter_generator_->addSelectionInfo(physical_selection, select_index);
@@ -679,12 +688,71 @@ void ExecutionGenerator::convertFilterJoin(const P::FilterJoinPtr &physical_plan
       std::piecewise_construct,
       std::forward_as_tuple(physical_plan),
       std::forward_as_tuple(probe_relation_info->producer_operator_index,
-                            probe_relation_info->relation));
+                            probe_relation_info->relation,
+                            probe_relation_info->output_destination_index));
 
   DCHECK(lip_filter_generator_ != nullptr);
   lip_filter_generator_->addFilterJoinInfo(physical_plan,
                                            build_filter_operator_index);
 }
+
+namespace {
+
+bool areSamePartitionSchemeHeaders(const PartitionSchemeHeader &lhs_partition_header,
+                                   const CatalogRelationSchema &lhs_scheme,
+                                   const PartitionSchemeHeader &rhs_partition_header,
+                                   const CatalogRelationSchema &rhs_scheme) {
+  if (lhs_partition_header.getPartitionType() != rhs_partition_header.getPartitionType()) {
+    return false;
+  }
+
+  if (lhs_partition_header.getNumPartitions() != rhs_partition_header.getNumPartitions()) {
+    return false;
+  }
+
+  // Check whether the underlying types in CatalogAttribute are the same.
+  if (!lhs_scheme.getAttributeById(lhs_partition_header.getPartitionAttributeId())->getType().equals(
+       rhs_scheme.getAttributeById(rhs_partition_header.getPartitionAttributeId())->getType())) {
+    return false;
+  }
+
+  switch (lhs_partition_header.getPartitionType()) {
+    case PartitionSchemeHeader::PartitionType::kHash:
+      return true;
+    case PartitionSchemeHeader::PartitionType::kRange: {
+      const vector<TypedValue> &lhs_ranges =
+          static_cast<const RangePartitionSchemeHeader&>(lhs_partition_header).getPartitionRangeBoundaries();
+      const vector<TypedValue> &rhs_ranges =
+          static_cast<const RangePartitionSchemeHeader&>(rhs_partition_header).getPartitionRangeBoundaries();
+
+      return lhs_ranges == rhs_ranges;
+    }
+  }
+
+  return false;
+}
+
+
+// Note that this method will be deprecated once the partition scheme header
+// supports multiple partition attributes.
+size_t chooseBestRepartitionAttributeIndex(const CatalogRelationStatistics &stats,
+                                           const vector<attribute_id> &join_attributes) {
+  size_t chose_attr_index = static_cast<size_t>(-1);
+  size_t chose_attr_num_distinct_values = 0;
+
+  for (std::size_t i = 0; i < join_attributes.size(); ++i) {
+    const attribute_id attr = join_attributes[i];
+    if (stats.hasNumDistinctValues(attr) &&
+        stats.getNumDistinctValues(attr) > chose_attr_num_distinct_values) {
+      chose_attr_index = i;
+      chose_attr_num_distinct_values = stats.getNumDistinctValues(attr);
+    }
+  }
+
+  return (chose_attr_index != static_cast<size_t>(-1)) ? chose_attr_index : 0;
+}
+
+}  // namespace
 
 void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
   // HashJoin is converted to three operators:
@@ -696,13 +764,10 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
   std::vector<attribute_id> probe_attribute_ids;
   std::vector<attribute_id> build_attribute_ids;
 
-  std::size_t build_cardinality =
-      cost_model_for_hash_join_->estimateCardinality(build_physical);
-
   bool any_probe_attributes_nullable = false;
   bool any_build_attributes_nullable = false;
 
-  const std::vector<E::AttributeReferencePtr> &left_join_attributes =
+  std::vector<E::AttributeReferencePtr> left_join_attributes =
       physical_plan->left_join_attributes();
   for (const E::AttributeReferencePtr &left_join_attribute : left_join_attributes) {
     const CatalogAttribute *probe_catalog_attribute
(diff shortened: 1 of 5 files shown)


zuyu:
Assigned to @hbdeshmukh.

query_optimizer/ExecutionGenerator.hpp
```diff
@@ -135,9 +135,11 @@ class ExecutionGenerator {
    */
   struct CatalogRelationInfo {
     CatalogRelationInfo(const QueryPlan::DAGNodeIndex producer_operator_index_in,
-                        const CatalogRelation *relation_in)
+                        const CatalogRelation *relation_in,
+                        const QueryContext::insert_destination_id output_destination_index_in)
```
hbdeshmukh: What do the ``_in`` suffix mean?

query_optimizer/tests/execution_generator/Partition.test
```diff
@@ -15,19 +15,35 @@
 # specific language governing permissions and limitations
 # under the License.
 
-CREATE TABLE foo (id INT NULL,
-                  name CHAR(20))
+CREATE TABLE dim (id INT NULL,
```
hbdeshmukh: Optional, but it may be easier to read if the relation is named as ``dim_4_hash_partitions``.

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -679,13 +688,72 @@ void ExecutionGenerator::convertFilterJoin(const P::FilterJoinPtr &physical_plan
...
+                            probe_relation_info->output_destination_index));
 
   DCHECK(lip_filter_generator_ != nullptr);
   lip_filter_generator_->addFilterJoinInfo(physical_plan,
                                            build_filter_operator_index);
 }
 
+namespace {
+
+bool areSamePartitionSchemeHeaders(const PartitionSchemeHeader &lhs_partition_header,
```
hbdeshmukh: These functions in the anonymous namespace may better be defined in some other source file. This cpp file is getting monstrously big, and any efforts to reduce its size should be welcome.

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -679,13 +688,72 @@ void ExecutionGenerator::convertFilterJoin(const P::FilterJoinPtr &physical_plan
...
+  switch (lhs_partition_header.getPartitionType()) {
+    case PartitionSchemeHeader::PartitionType::kHash:
+      return true;
+    case PartitionSchemeHeader::PartitionType::kRange: {
+      const vector<TypedValue> &lhs_ranges =
+          static_cast<const RangePartitionSchemeHeader&>(lhs_partition_header).getPartitionRangeBoundaries();
+      const vector<TypedValue> &rhs_ranges =
+          static_cast<const RangePartitionSchemeHeader&>(rhs_partition_header).getPartitionRangeBoundaries();
+
+      return lhs_ranges == rhs_ranges;
```
hbdeshmukh: Can you help me understand how the comparison of two vector references will produce the desired result in this case?

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -679,13 +688,72 @@ void ExecutionGenerator::convertFilterJoin(const P::FilterJoinPtr &physical_plan
...
+
+  return false;
+}
+
+
+// Note that this method will be deprecated once the partition scheme header
+// supports multiple partition attributes.
+size_t chooseBestRepartitionAttributeIndex(const CatalogRelationStatistics &stats,
+                                           const vector<attribute_id> &join_attributes) {
+  size_t chose_attr_index = static_cast<size_t>(-1);
```
hbdeshmukh: This would assign ``chose_attr_index`` the value of ``MAX_INT``. Why do we want that?

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -679,13 +688,72 @@ void ExecutionGenerator::convertFilterJoin(const P::FilterJoinPtr &physical_plan
...
+  for (std::size_t i = 0; i < join_attributes.size(); ++i) {
+    const attribute_id attr = join_attributes[i];
+    if (stats.hasNumDistinctValues(attr) &&
+        stats.getNumDistinctValues(attr) > chose_attr_num_distinct_values) {
+      chose_attr_index = i;
+      chose_attr_num_distinct_values = stats.getNumDistinctValues(attr);
+    }
+  }
+
+  return (chose_attr_index != static_cast<size_t>(-1)) ? chose_attr_index : 0;
```
hbdeshmukh: I am unable to understand the point of this function. Why do we need to find the "best repartition attribute"? Assuming a join like ``R.a = S.b``, if ``R`` is partitioned on some other attribute ``c``, we would like to repartition it based on ``a``, so the only choice for an attribute on which we should perform the repartitioning is ``a``, isn't it? What's the decision got do do with the CatalogRelation stats and the number of distinct values of the attribute?

query_optimizer/ExecutionGenerator.hpp
zuyu: It is used to name an argument in the constructor for a struct.

query_optimizer/ExecutionGenerator.cpp
zuyu: The comparison result checks every value in both vectors to ensure whether two have the same range partitions.

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -740,6 +805,202 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
...
+            }
+            break;
+          }
+        }
+        break;
+      case PartitionSchemeHeader::PartitionType::kHash: {
+        const attribute_id build_partition_attr = build_partition_scheme_header.getPartitionAttributeId();
+        if (find(build_attribute_ids.begin(), build_attribute_ids.end(), build_partition_attr) !=
+                build_attribute_ids.end()) {
+          // BuildRelation has an useful partition.
```
hbdeshmukh: Grammar: an -> a

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -740,6 +805,202 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
...
+
+    switch (build_partition_scheme_header.getPartitionType()) {
+      case PartitionSchemeHeader::PartitionType::kRange:
+        build_needs_repartition = true;
+
+        switch (probe_partition_scheme_header.getPartitionType()) {
+          case PartitionSchemeHeader::PartitionType::kRange:
+            probe_needs_repartition = true;
+            break;
+          case PartitionSchemeHeader::PartitionType::kHash: {
```
hbdeshmukh: Should be good to add a default case here.

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -740,6 +805,202 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
...
+            const attribute_id probe_partition_attr = probe_partition_scheme_header.getPartitionAttributeId();
+            if (find(probe_attribute_ids.begin(), probe_attribute_ids.end(), probe_partition_attr) !=
+                    probe_attribute_ids.end()) {
+              needs_swap = true;
```
hbdeshmukh: Why do we need to swap here? There's no cardinality check performed here unlike line 863.

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -740,6 +805,202 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
...
+      case PartitionSchemeHeader::PartitionType::kHash: {
+        const attribute_id build_partition_attr = build_partition_scheme_header.getPartitionAttributeId();
+        if (find(build_attribute_ids.begin(), build_attribute_ids.end(), build_partition_attr) !=
+                build_attribute_ids.end()) {
+          // BuildRelation has an useful partition.
+          switch (probe_partition_scheme_header.getPartitionType()) {
+            case PartitionSchemeHeader::PartitionType::kRange:
+              probe_needs_repartition = true;
+              break;
+            case PartitionSchemeHeader::PartitionType::kHash: {
```
hbdeshmukh: Again, a default case should make the code more extensible.

zuyu: I intend not to add a default case, so that once a new partition scheme is added, we will have compile errors to force to revisit how the new partition scheme fits in this method.

query_optimizer/ExecutionGenerator.cpp
zuyu: In a hash join, if the build relation uses range partition, while the probe uses hash partition, we swap two relations, and then repartition the probe (aka, the original build) relation.

query_optimizer/ExecutionGenerator.cpp
zuyu: In a hash join, if both relations are using range partitions, we need to repartition both relations. Here we use this method to pick up the best single attribute for the new hash repartition. By best, we mean the maximum number of distinct values to minimize the hash collision.

Finally, once we support multiple partition attributes, we don't need this method.

query_optimizer/ExecutionGenerator.cpp
zuyu: That's just an initial value.

query_optimizer/ExecutionGenerator.cpp
zuyu: Unless these functions are useful to other files, we keep them in the same file as the caller.

hbdeshmukh: I had a discussion with @jianqiao on this issue. He thinks that the partitioning related code should be moved to different functions at least. That way the overall code is easier to read and is also extensible. He's fine with the anonymous namespace arrangement. 

An example is all the new code in ``convertHashJoin`` can be moved to separate function(s) and then called from ``convertHashJoin``.

query_optimizer/ExecutionGenerator.cpp
hbdeshmukh: Are you sure that the compiler will throw an error saying that we didn't handle all the cases in the switch statement? I would recommend a safe approach where we add a default case that throws a fatal error.

query_optimizer/ExecutionGenerator.cpp
hbdeshmukh: I am not sure if I agree. Consider the following counterexample. 
Original build table - cardinality 10k, range partitioned. 
Original probe table - cardinality 1M, hash partitioned. 

By the above logic, you will swap the build and probe relations, thereby building a giant hash table for 1M entries. You are going to repartition the new probe relation anyway, so that cost isn't avoided.

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -740,6 +805,202 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
...
+              } else {
+                probe_needs_repartition = true;
+              }
+              break;
+            }
+          }
+        }
+        break;
+      }
+    }
```
hbdeshmukh: I would highly recommend two things for the switch statement:
1. Add several comments in between to describe what's happening. I felt lost while reading it.
2. Moving the switch statement to a separate method.

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -740,6 +805,202 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
...
+      S::InsertDestination *build_insert_destination_proto =
+          query_context_proto_->mutable_insert_destinations(build_relation_info->output_destination_index);
+
+      build_insert_destination_proto->set_insert_destination_type(S::InsertDestinationType::PARTITION_AWARE);
+      build_insert_destination_proto->MutableExtension(S::PartitionAwareInsertDestination::partition_scheme)
+        ->MergeFrom(build_repartition_scheme->getProto());
+
+      CatalogRelation *mutable_build_relation =
+          catalog_database_->getRelationByIdMutable(build_relation->getID());
+      mutable_build_relation->setPartitionScheme(build_repartition_scheme.release());
```
hbdeshmukh: May be I missed some of your older PRs, but can you explain how we will physically repartition the relations?

hbdeshmukh:
One case which is currently not considered is the following:

Imagine a join between two relations. The probe side is a result of a selection on a base relation. The probe's base relation is hash partitioned on attribute ``a``. There are two cases:
1. In the projection, the attribute ``a`` is left out, i.e. we don't project it. 
2. We project the attribute ``a``. 

If you look at the code of ``convertSelection``, the output of the selection has no partitioning scheme. In case two, we can carry forward the partitioning scheme from the base table to the output of the selection. That way, you can avoid the repartitioning step for the probe side.

To summarize, you need to add a check in ``convertSelection`` that can identify case 1 and case 2 mentioned earlier.

query_optimizer/ExecutionGenerator.cpp
zuyu: I get your point, but the partition number is set by us. So we won't have such insane number, but I could add a TODO here.

query_optimizer/ExecutionGenerator.cpp
zuyu: Yes, it is `-Wswitch` and then becomes an error for us.

query_optimizer/ExecutionGenerator.cpp
pateljm: +1 for the suggestion to move the partition code to a different file. It is (related) but separate module.

query_optimizer/ExecutionGenerator.cpp
zuyu: We modify the existing `SelectOperator` to have a new `PartitionAwareInsertDestination` for repartition, if the input relation is not a base table. Otherwise, we will create one, but not part of this PR.

query_optimizer/ExecutionGenerator.cpp
hbdeshmukh: It's clearly not an insane number. I said **cardinality** and not **number of partitions**. SSB Lineorder table for SF100 has a cardinality of 600M, so 1M is not insane by any means.

zuyu: Now I got your point. I'll do another swap if needed, so that the build side has a less cardinality. Thanks!

zuyu:
Close it temporally.

PR #183, opened 2017-02-07
Adds regex to specify 16.+ versions of Darwin

cramja:
Before this change, only one recent version of Darwin was correctly
having the compiler flags modified to allow for deprecated syscalls.
After this change, all versions of Darwin 16+ should correctly apply the
cxx flags. This is still a somewhat temporary fix because the real
issue (replacing the deprecated calls) has not been addressed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -299,7 +299,7 @@ else()
   # builds to fail. As a short-term workaround for now, we turn off deprecated
   # warnings so that they do not cause build failures anymore.
   # TODO: Remove this workaround by fixing the protobuf_cmake and glog_cmake.
-  if (${CMAKE_SYSTEM} MATCHES "Darwin-16.1.0")
+  if (${CMAKE_SYSTEM} MATCHES "Darwin-16.[0-9]*.[0-9]*")
     if (${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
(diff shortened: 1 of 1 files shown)


cramja:
BTW this is so we can build on Mac. @pateljm 

I'm still finding that I must specify `cmake -DUSE_TCMALLOC=FALSE ..`. Then it builds successfully on Mac.

cramja:
@jianqiao can I get a +1 when travis comes back?

ps
This doesn't solve the TCMALLOC issue. I tried using dynamic->static libs, and building perftools standalone. It compiles fine if I do it outside of CMAKE... so therefore it seems to be an issue with `ExternalProject`. 

Did TCMALLOC ever work with Mac? Maybe this has to do with the 3rd party change, or not, I don't remember.

zuyu:
I sometimes encountered `tcmalloc` build issues on Mac, but it worked for a while before the third party changes.

cramja:
What was your solution to tc malloc issue on mac, just ignore it?

zuyu:
So far it is only the problem for Mac, so I disabled `tcmalloc` by `-DUSE_TCMALLOC=0` in `cmake` configuration.

cramja:
Merged, closing

zuyu:
Next time, before merging, sync again your local branch with the master, and this PR will automatically merged, instead of manually closed.

pateljm:
Thanks @cramja !! +1

PR #185, opened 2017-02-09
Fuse Aggregate with HashLeftOuterJoin to accelerate evaluation

jianqiao:
This PR fuses `Aggregate` with `HashJoin` (left outer) into the `CrossReferenceCoalesceAggregate` node to enable fast-path evaluation for a class of queries.

Here we briefly describe the semantics of `CrossReferenceCoalesceAggregate`: Let `L` be a table with PRIMARY KEY `u`. Let `R` be a table with FOREIGN KEY `x` referring to `L(u)`. Then `CrossReferenceCoalesceAggregate` represents a common class of analytical queries that
> For each u in L, COUNT/SUM the records in R that correspond to u (i.e. those records satisfying R.x = L.u). In the case that there is no record for u in R, use 0 as the result value.

That is, `CrossReferenceCoalesceAggregate` represents queries in the following forms:
```
-- Form 1 --
SELECT u, COALESCE(count, 0), COALESCE(sum, 0)
FROM L LEFT OUTER JOIN (
  SELECT x, COUNT(*) AS count, SUM(y) AS sum
  FROM R
  GROUP BY x) Rt ON L.u = Rt.x;
```
or (COUNT only) 
```
-- Form 2 --
SELECT u, COUNT(x) AS count
FROM L LEFT OUTER JOIN R ON L.u = R.x
GROUP BY u;
```
where `L`, `R`, `u`, `v` and `SUM(...)/COUNT(...)` are arguments to `CrossReferenceCoalesceAggregate`.

The fast-path evaluation is that, we first use `L.u` to build the `existence_map` for a `CollisionFreeVectorTable` which is then used to perform aggregation on `R` grouped by `R.x`. At the time of aggregation finalization, any `existence_map` entry that is set by `L.u` but not updated with `R`'s records will have aggregation result as `0` by default -- this implicitly achieves the left outer join + coalesce 0 semantics.

This optimization improves TPC-H Q13's performance from ~15.5s to ~3s on a cloudlab machine with scale factor 100.

### DIFF
--- a/query_optimizer/physical/CrossReferenceCoalesceAggregate.hpp
+++ b/query_optimizer/physical/CrossReferenceCoalesceAggregate.hpp
@@ -1,0 +1,232 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ **/
+
+#ifndef QUICKSTEP_QUERY_OPTIMIZER_PHYSICAL_CROSS_REFERENCE_COALESCE_AGGREGATE_HPP_
+#define QUICKSTEP_QUERY_OPTIMIZER_PHYSICAL_CROSS_REFERENCE_COALESCE_AGGREGATE_HPP_
+
+#include <cstddef>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "query_optimizer/OptimizerTree.hpp"
+#include "query_optimizer/expressions/Alias.hpp"
+#include "query_optimizer/expressions/AttributeReference.hpp"
+#include "query_optimizer/expressions/ExpressionUtil.hpp"
+#include "query_optimizer/expressions/Predicate.hpp"
+#include "query_optimizer/physical/Physical.hpp"
+#include "query_optimizer/physical/PhysicalType.hpp"
+#include "utility/Macros.hpp"
+
+#include "glog/logging.h"
+
+namespace quickstep {
+namespace optimizer {
+namespace physical {
+
+/** \addtogroup OptimizerLogical
+ *  @{
+ */
+
+class CrossReferenceCoalesceAggregate;
+typedef std::shared_ptr<const CrossReferenceCoalesceAggregate> CrossReferenceCoalesceAggregatePtr;
+
+/**
+ * @brief A physical node that fuses a HashJoin with an Aggregate to enable
+ *        fast-path execution.
+ *
+ * Below we briefly describe the semantics of this physical node.
+ *
+ * Let L be a table with PRIMARY KEY u. Let R be a table with FOREIGN KEY x
+ * referring to L(u). Then CrossReferenceCoalesceAggregate represents a common
+ * class of analytical queries that
+ * - For each u in L, COUNT/SUM the records in R that correspond to u (i.e.
+ *   those records satisfying R.x = L.u).
+ *   In the case that there is no record for u in R, use 0 as the result value.
+ *
+ * And we have the mapping:
+ *   L -> left_child_
+ *   R -> right_child_
+ *   u -> left_join_attributes_
+ *   x -> right_join_attributes_
+ *   COUNT/SUM -> aggregate_expressions_
+ */
+class CrossReferenceCoalesceAggregate : public Physical {
+ public:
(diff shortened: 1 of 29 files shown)


pateljm:
LGTM. Merging.

query_optimizer/physical/CrossReferenceCoalesceAggregate.hpp
```diff
@@ -0,0 +1,232 @@
...
+typedef std::shared_ptr<const CrossReferenceCoalesceAggregate> CrossReferenceCoalesceAggregatePtr;
+
+/**
+ * @brief A physical node that fuses a HashJoin with an Aggregate to enable
+ *        fast-path execution.
+ *
+ * Below we briefly describe the semantics of this physical node.
+ *
+ * Let L be a table with PRIMARY KEY u. Let R be a table with FOREIGN KEY x
+ * referring to L(u). Then CrossReferenceCoalesceAggregate represents a common
```
zuyu: Should be `L.u`?

query_optimizer/physical/CrossReferenceCoalesceAggregate.hpp
```diff
@@ -0,0 +1,232 @@
...
+
+  // TODO(jianqiao): For the left child, support filter predicate fusing and
+  // attachment of LIPFilters.
+  PhysicalPtr left_child_;
+  PhysicalPtr right_child_;
+  std::vector<expressions::AttributeReferencePtr> left_join_attributes_;
+  std::vector<expressions::AttributeReferencePtr> right_join_attributes_;
+  expressions::PredicatePtr right_filter_predicate_;
+  std::vector<expressions::AliasPtr> aggregate_expressions_;
+  std::size_t group_by_key_value_range_;
```
zuyu: Would above data members change? Otherwise, we could mark `const`.

relational_operators/BuildAggregationExistenceMapOperator.cpp
```diff
@@ -0,0 +1,196 @@
...
+      ++num_workorders_generated_;
+    }
+    return done_feeding_input_relation_;
+  }
+}
+
+serialization::WorkOrder* BuildAggregationExistenceMapOperator
+    ::createWorkOrderProto(const block_id block) {
+  serialization::WorkOrder *proto = new serialization::WorkOrder;
+  proto->set_work_order_type(serialization::BUILD_LIP_FILTER);
```
zuyu: This is a bug: we need to use the new type called `BUILD_AGGREGATION_EXISTENCE_MAP` here.

relational_operators/BuildAggregationExistenceMapOperator.hpp
```diff
@@ -0,0 +1,177 @@
...
+    input_relation_block_ids_.push_back(input_block_id);
+  }
+
+ private:
+  serialization::WorkOrder* createWorkOrderProto(const block_id block);
+
+  const CatalogRelation &input_relation_;
+  const attribute_id build_attribute_;
+  const bool input_relation_is_stored_;
+  std::vector<block_id> input_relation_block_ids_;
```
zuyu: Please move this line before `num_workorders_generated_`, so that all constants are together.

relational_operators/BuildAggregationExistenceMapOperator.hpp
```diff
@@ -0,0 +1,177 @@
...
+  ~BuildAggregationExistenceMapWorkOrder() override {}
+
+  void execute() override;
+
+ private:
+  const CatalogRelationSchema &input_relation_;
+  const block_id build_block_id_;
+  const attribute_id build_attribute_;
+  AggregationOperationState *state_;
+
```
zuyu: Move this empty line before `state_`, so that all constants are together.

relational_operators/WorkOrder.proto
```diff
@@ -44,6 +44,7 @@ enum WorkOrderType {
   UPDATE = 20;
   WINDOW_AGGREGATION = 21;
   DESTROY_AGGREGATION_STATE = 22;
+  BUILD_AGGREGATION_EXISTENCE_MAP = 23;
```
zuyu: Please sort the type in the alphabet order (similarly to `BuildAggregationExistenceMapWorkOrder` below), and add the following comment above `enum WorkOrderType {`:

```
// Next tag: 24.
enum WorkOrderType {
```

Recall that this is how Google deals with changing proto definitions. And `Next tag` is the next available number to assign to a new type. After adding the new type, we need to increase `Next tag` by one.

relational_operators/WorkOrderFactory.cpp
```diff
@@ -91,6 +92,19 @@ WorkOrder* WorkOrderFactory::ReconstructFromProto(const serialization::WorkOrder
...
+      LOG(INFO) << "Creating BuildAggregationExistenceMapWorkOrder in Shiftboss " << shiftboss_index;
+
+      return new BuildAggregationExistenceMapWorkOrder(
+          proto.query_id(),
+          catalog_database->getRelationSchemaById(
+              proto.GetExtension(serialization::BuildAggregationExistenceMapWorkOrder::relation_id)),
+          proto.GetExtension(serialization::BuildAggregationExistenceMapWorkOrder::build_block_id),
+          proto.GetExtension(serialization::BuildAggregationExistenceMapWorkOrder::build_attribute),
+          query_context->getAggregationState(
+              proto.GetExtension(serialization::AggregationWorkOrder::aggr_state_index)),
```
zuyu: This is a bug: replace `AggregationWorkOrder` with `BuildAggregationExistenceMapWorkOrder`.

PR #188, opened 2017-02-14
Adds marcs ssh key to KEYS

cramja:
For release. I looked at apache KUDU and IMPALA and they both do not keep the KEYS file in the root of their git project. I wonder if we only need to keep this in the svn repo holding the releases.


cramja:
LGTM, merging

PR #189, opened 2017-02-21
patches for missed linenoise changes

cramja:
This fixes an issue where newlines were not being given to quickstep's parser. These changes existed previously in quickstep but were not carried over in #166

### DIFF
--- a/third_party/download_and_patch_prerequisites.sh
+++ b/third_party/download_and_patch_prerequisites.sh
@@ -89,7 +89,11 @@ do
 done
 
 # Apply patches now.
+
+# Apply linenoise patch
 cp ${PATCH_DIR}/linenoise/CMakeLists.txt ${THIRD_PARTY_SRC_DIR}/linenoise
+patch ${THIRD_PARTY_SRC_DIR}/linenoise/linenoise.h ${PATCH_DIR}/linenoise/linenoise.h.patch
+patch ${THIRD_PARTY_SRC_DIR}/linenoise/linenoise.c ${PATCH_DIR}/linenoise/linenoise.c.patch
 
 # Apply gflags patch.
 echo "Patching for gflags:"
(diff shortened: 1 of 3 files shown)


cramja:
@jianqiao This does not fix the bug with `\analyze`. That appears to be something with how the analyze command was written.

jianqiao:
LGTM. Thanks @cramja for the fix!

zuyu:
@cramja The patch has trailing whitespaces.

zuyu:
@cramja Did this PR fix the issue that the cli shell does not work with a input file with multiple queries?

cramja:
@zuyu yes it fixed the multiple queries issue.

zuyu:
@cramja In addition to `\analyze`, it seems that other commands like `\d` and `\dt` also do not work.

cramja:
@zuyu I cannot recreate the issue you are describing.

```
Starting Quickstep with 8 worker thread(s) and a 11.20 GB buffer pool.
--num_workers is 8, but only specified 0 CPU affinities with --worker_affinities. 8 workers will be unaffinitized.
quickstep> \d
quickstep> \dt
quickstep> \dt foo
ERROR:  Unrecognized relation foo (1 : 5)
\dt foo
    ^
quickstep> create table foo (id int);
Time: 0.205 ms
quickstep> \d
       List of relations

 Name   | Type  | Blocks
+-------+-------+---------+
 foo    | table | 0

quickstep> \dt foo
       List of relations

 Name   | Type  | Blocks
+-------+-------+---------+
 foo    | table | 0

quickstep> \d foo
 Table "foo"
 Column | Type
+-------+-------+
 id     | Int
quickstep> \d foobar
ERROR:  Unrecognized relation foobar (1 : 4)
\d foobar
   ^
quickstep> \d
       List of relations

 Name   | Type  | Blocks
+-------+-------+---------+
 foo    | table | 0

quickstep> quit;
```

zuyu:
Never mind. It works after resetting the third party and recompiling.

PR #190, opened 2017-02-22
Use BitVectorExactFilter as LIPFilter implementation when applicable

jianqiao:
In the previous exact-filter PR, it was not by default use `BitVectorExactFilter` as the LIPFilter implementation when applicable -- `BitVectorExactFilter` was only used for executing `FilterJoin` nodes.

This PR is a quick update that uses the `BitVectorExactFilter` when applicable.


hakanmemisoglu:
It looks good to me. Merging.

PR #193, opened 2017-02-24
Use partitioned aggregation for single-function DISTINCT aggregation.

jianqiao:
This PR uses the partitioned aggregation routine for parallelizing DISTINCT aggregation. It improves the performance of TPC-H Q16 from ~10s to ~2s with scale factor 100 on CloudLab machine.


zuyu:
@jianqiao Please fix the build failures using gcc. See [here](https://travis-ci.org/apache/incubator-quickstep/builds/205062637) for more info.

PR #196, opened 2017-02-27
Fix gcc build error with PackedPayloadHashTable 

jianqiao:
This PR fixes the gcc build error (missing `this` pointer for calling a method in a capture-all-by-reference lambda function).

Note: Modified `travis.yml` for testing, do not merge yet.


zuyu:
I think this fix should work!

cramja:
+1

PR #198, opened 2017-02-28
patch to fix gcc compile error gflags

cramja:
This fix adds a patch to the third party library `gflags` so that compilation will work with certain versions of GCC which were previously erroring on a warn->error for an unsigned to signed conversion.

@jianqiao


jianqiao:
Great! Thanks Marc @cramja.

zuyu:
@cramja If I remembered correctly, we disable some waring->error flags for building third party libraries. So for this issue, I suggest to disable certain cxx flags, instead of adding a patch.

cramja:
The patch fixes arguably bad style on the part of the 3rd party library authors, so I do not think there's reason to revert the patch. However, in the future it would be much easier to just change the compiler flags like you suggested. +1

zuyu:
It is more about maintainability when we try to upgrade the third party while our patch may not work in a new release.

PR #201, opened 2017-03-01
Fix a bug with SelectOperator

jianqiao:
This PR fixes a bug that `SELECT ... LIMIT n;` may output more than `n` rows.

The cause is that `SortMergeRunOperator` creates two relations, where one relation is the output relation to a `SelectOperator` and the other is a _temporary_ relation. However, the generated blocks of the temporary relation are also fed into the `SelectOperator` via a _DataPipelineMessage_ -- thus we need to check that the input block to a `SelectOperator` is actually from its input relation.


hbdeshmukh:
LGTM, merging.

PR #204, opened 2017-03-03
Cmake changes for default build

cramja:
This fixes two annoying build issues

1. If the CMAKEBUILDTYPE is not set by the user, it previously defaulted to none-type which caused one of our tests to fail. It is now set to RelWithDebInfo if no build type is set. Code take from [link](https://github.com/miloyip/rapidjson/issues/319)
2. When building on Mac, the default setting of TCMALLOC would cause a compiler error. We fix this by disabling TCMALLOC on Mac

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,11 @@ include(CheckCCompilerFlag)
 include(CheckCXXCompilerFlag)
 include(CheckCXXSourceCompiles)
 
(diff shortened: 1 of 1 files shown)


cramja:
@hbdeshmukh this should go in before RC6

hbdeshmukh:
Hi @cramja Can you please rebase this branch? I will merge it thereafter.

PR #210, opened 2017-03-10
Fixed a bug in admitQueries.

zuyu:
Assigned to @hbdeshmukh.

This PR fixed the bug in `admitQueries` that if an `AdmitRequestMessage` contains a number of queries that excesses that can be admitted to the system, the excessed queries would be lost.

Instead, now we add these overflowed queries to a waiting list so that they will be scheduled later.


hbdeshmukh:
Looks good to me. Will merge after travis finishes.

PR #217, opened 2017-03-18
QUICKSTEP-73: Multi-attribute PartitionSchemeHeader.

zuyu:
This PR added multi-attribute partitions support.

```
CREATE TABLE foo5 (attr1 INT, attr2 LONG, attr3 FLOAT, attr4 DOUBLE, attr5 CHAR(5),
                   attr6 VARCHAR(4))
PARTITION BY HASH(attr1, attr2, attr3, attr4, attr5, attr6) PARTITIONS 4;
```

### DIFF
--- a/catalog/PartitionSchemeHeader.cpp
+++ b/catalog/PartitionSchemeHeader.cpp
@@ -20,10 +20,12 @@
 #include "catalog/PartitionSchemeHeader.hpp"
 
 #include <cstddef>
+#include <unordered_set>
 #include <utility>
 #include <vector>
 
 #include "catalog/Catalog.pb.h"
+#include "catalog/CatalogTypedefs.hpp"
 #include "types/Type.hpp"
 #include "types/Type.pb.h"
 #include "types/TypeFactory.hpp"
@@ -32,16 +34,30 @@
 
 #include "glog/logging.h"
 
+using std::move;
+using std::size_t;
+using std::vector;
+
 namespace quickstep {
 
 PartitionSchemeHeader::PartitionSchemeHeader(const PartitionType type,
                                              const std::size_t num_partitions,
-                                             const attribute_id attr_id)
+                                             PartitionAttributeIds &&attr_ids)  // NOLINT(whitespace/operators)
     : partition_type_(type),
       num_partitions_(num_partitions),
-      partition_attribute_id_(attr_id) {
+      partition_attribute_ids_(move(attr_ids)) {
   DCHECK_GT(num_partitions, 0u);
-  DCHECK_GE(attr_id, 0);
+
+#ifdef QUICKSTEP_DEBUG
+  // Ensure that no duplicated partition attributes exist.
+  std::unordered_set<attribute_id> partition_attribute_ids;
+  for (const attribute_id attr_id : partition_attribute_ids_) {
+    DCHECK_NE(attr_id, kInvalidCatalogId);
+
+    CHECK_EQ(0u, partition_attribute_ids.count(attr_id));
+    partition_attribute_ids.insert(attr_id);
+  }
+#endif
 }
 
 bool PartitionSchemeHeader::ProtoIsValid(
@@ -60,10 +76,12 @@ bool PartitionSchemeHeader::ProtoIsValid(
     case serialization::PartitionSchemeHeader::HASH:
       return true;
     case serialization::PartitionSchemeHeader::RANGE: {
-      const std::size_t num_ranges =
-          proto.ExtensionSize(serialization::RangePartitionSchemeHeader::partition_range_boundaries);
-      return num_ranges == proto.num_partitions() - 1 &&
-             proto.HasExtension(serialization::RangePartitionSchemeHeader::partition_attr_type);
+      return proto.HasExtension(serialization::RangePartitionSchemeHeader::partition_range_boundaries) &&
+             proto.ExtensionSize(serialization::RangePartitionSchemeHeader::partition_range_boundaries) ==
+                 static_cast<int>(proto.num_partitions() - 1) &&
+             proto.HasExtension(serialization::RangePartitionSchemeHeader::partition_attr_types) &&
+             proto.ExtensionSize(serialization::RangePartitionSchemeHeader::partition_attr_types) ==
+                 proto.partition_attribute_ids_size();
     }
     default:
       // Partition type is unknown.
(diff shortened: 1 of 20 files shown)


hbdeshmukh:
Hi @zuyu 

Looks like a compilation error is causing travis to fail.

catalog/PartitionSchemeHeader.cpp
```diff
@@ -32,16 +34,29 @@
...
-                                             const attribute_id attr_id)
+                                             PartitionAttributeIds &&attr_ids)  // NOLINT(whitespace/operators)
     : partition_type_(type),
       num_partitions_(num_partitions),
-      partition_attribute_id_(attr_id) {
+      partition_attribute_ids_(move(attr_ids)) {
   DCHECK_GT(num_partitions, 0u);
-  DCHECK_GE(attr_id, 0);
+
+#ifdef QUICKSTEP_DEBUG
```
hbdeshmukh: Optional, but a couple of lines regarding what you are trying to do here should be helpful.

catalog/PartitionSchemeHeader.cpp
```diff
@@ -77,32 +95,44 @@ PartitionSchemeHeader* PartitionSchemeHeader::ReconstructFromProto(
...
+      return new RangePartitionSchemeHeader(proto.num_partitions(),
+                                            move(partition_attribute_ids),
+                                            move(attr_types),
+                                            move(partition_ranges));
     }
     default:
       LOG(FATAL) << "Invalid partition scheme header.";
   }
+
+  return nullptr;
```
hbdeshmukh: Isn't this return statement redundant?

catalog/PartitionSchemeHeader.hpp
```diff
@@ -48,6 +50,13 @@ class Type;
  **/
 class PartitionSchemeHeader {
  public:
+  // A vector for partitioning catalog attribute .
+  typedef std::vector<attribute_id> PartitionAttributeIds;
+
+  // A boundary for multi-attribute partitions.
+  // PartitionValues.size() should be equal to PartitionAttributeIds.size().
+  typedef std::vector<TypedValue> PartitionValues;
```
hbdeshmukh: A clearer and more informative name should be ``PartitionBoundaryValues``.

catalog/PartitionSchemeHeader.hpp
```diff
@@ -180,16 +190,17 @@ class HashPartitionSchemeHeader : public PartitionSchemeHeader {
    * @brief Calulate the partition id into which the attribute value
    *        should be inserted.
    *
-   * @param value_of_attribute The attribute value for which the
-   *                           partition id is to be determined.
+   * @param value_of_attributes The attribute value for which the
```
hbdeshmukh: Nitpick - attribute values instead of attribute value.

catalog/PartitionSchemeHeader.hpp
```diff
@@ -180,16 +190,17 @@ class HashPartitionSchemeHeader : public PartitionSchemeHeader {
...
+   *                            partition id is to be determined.
    * @return The partition id of the partition for the attribute value.
    **/
   partition_id getPartitionId(
-      const TypedValue &value_of_attribute) const override {
+      const PartitionValues &value_of_attributes) const override {
```
hbdeshmukh: In the type definition of ``PartitionValues``, you said it is for representing the boundary for partition values. That doesn't seem true here. Why not just use the vector of ``TypedValue`` instead?

catalog/PartitionSchemeHeader.hpp
```diff
@@ -249,20 +269,22 @@ class RangePartitionSchemeHeader : public PartitionSchemeHeader {
    * @brief Calulate the partition id into which the attribute value
    *        should be inserted.
    *
-   * @param value_of_attribute The attribute value for which the
-   *                           partition id is to be determined.
+   * @param value_of_attributes The attribute value for which the
```
hbdeshmukh: Nitpick - values instead of value.

catalog/PartitionSchemeHeader.hpp
```diff
@@ -288,22 +310,46 @@ class RangePartitionSchemeHeader : public PartitionSchemeHeader {
...
+    }
+  }
+
+  bool lessThan(const PartitionValues &lhs, const PartitionValues &rhs) const {
+    DCHECK_EQ(partition_attribute_ids_.size(), lhs.size());
+    DCHECK_EQ(partition_attribute_ids_.size(), rhs.size());
+
+    for (std::size_t attr_index = 0; attr_index < partition_attribute_ids_.size(); ++attr_index) {
+      if (less_unchecked_comparators_[attr_index]->compareTypedValues(lhs[attr_index], rhs[attr_index])) {
+        break;
```
hbdeshmukh: Do you mean to use ``continue`` here? The break will cause the control to come out of the for loop and we wouldn't have compared all the attributes, subsequently returning true. Am I missing something?

catalog/tests/PartitionScheme_unittest.cpp
```diff
@@ -343,15 +363,43 @@ TEST(PartitionSchemeHeaderTest, VarCharRangePartitionSchemeHeaderTest) {
...
-                             std::strlen(kSampleStrings[i]) + 1)));
+                  { TypedValue(kVarChar, kSampleStrings[i], std::strlen(kSampleStrings[i]) + 1) }));
   }
 }
 
+TEST(PartitionSchemeHeaderTest, MultiAttributeRangePartitionSchemeHeaderTest) {
+  vector<PartitionSchemeHeader::PartitionValues> partition_ranges;
+  // Partition boundaries are { 0, 0.00000 }, { 10, 10.0000 }, { 20, 20.00000 }
+  // Last partition can hold upto infinity.
+  // First partition can hold from { -infinity, -infinity } to { 0, -2^(-1074) }.
```
hbdeshmukh: Can you explain -2^(-1074)? I thought the first partition should be < {0, 0.0000} and > {-Inf, -Inf}.

catalog/PartitionSchemeHeader.hpp
hbdeshmukh: I think your matching definition is as follows:
(x1, x2, x3, ..., xk) < (y1, y2, y3, ..., yk) if (x1 < x2) || (x2, x3, ..., xk) < (y2, y3, ... yk). 

If that's correct, this should be clearly stated somewhere in the code.

catalog/PartitionSchemeHeader.cpp
zuyu: In [Travis CI](https://travis-ci.org/apache/incubator-quickstep/jobs/212429450#L1226), the compiler complains `-Werror=return-type` here.

catalog/PartitionSchemeHeader.hpp
zuyu: Sorry, the comment is not that clear. This represents the corresponding values for `PartitionAttributeIds`, and is also used to store the partition range boundaries for a range-based partition.

catalog/tests/PartitionScheme_unittest.cpp
```diff
@@ -611,6 +665,7 @@ TEST(PartitionSchemeTest, CheckRangePartitionSchemeSerialization) {
     EXPECT_EQ(blocks_in_part_scheme, blocks_in_part_scheme_from_proto);
   }
 }
+*/
```
hbdeshmukh: Is there some issue with the commented code?

query_optimizer/resolver/Resolver.cpp
```diff
@@ -724,6 +708,25 @@ const S::PartitionSchemeHeader* Resolver::resolvePartitionClause(
...
+  proto->set_num_partitions(partition_clause->num_partitions()->long_value());
+
+  std::unordered_set<attribute_id> unique_partition_attrs;
+  for (const ParseString &partition_attribute_name : partition_clause->attribute_name_list()) {
+    const attribute_id attr_id = GetAttributeIdFromName(create_table_statement.attribute_definition_list(),
+                                                        partition_attribute_name.value());
+    if (attr_id == kInvalidAttributeID) {
+      THROW_SQL_ERROR_AT(&partition_attribute_name)
+          << "The given attribute was not found.";
+    } else if (unique_partition_attrs.count(attr_id)) {
```
hbdeshmukh: For readability, can we change this to ``count > 0`` or use ``find`` instead?

query_optimizer/resolver/Resolver.cpp
```diff
@@ -724,6 +708,25 @@ const S::PartitionSchemeHeader* Resolver::resolvePartitionClause(
     THROW_SQL_ERROR_AT(partition_type_string) << "Unrecognized partition type: " << partition_type;
   }
 
+  proto->set_num_partitions(partition_clause->num_partitions()->long_value());
```
hbdeshmukh: Do we perform error checks for the following conditions?
1. Number of partitions = 0
2. No attribute specified in the partitioning attributes' list.

storage/InsertDestination.cpp
```diff
@@ -682,8 +685,11 @@ void PartitionAwareInsertDestination::bulkInsertTuplesWithRemappedAttributes(
     // set a bit in the appropriate TupleIdSequence.
     accessor->beginIteration();
     while (accessor->next()) {
-      TypedValue attr_val = accessor->getTypedValue(attribute_map[partition_attribute_id]);
-      partition_membership[partition_scheme_header_->getPartitionId(attr_val)]
+      PartitionSchemeHeader::PartitionValues values;
+      for (const attribute_id attr_id : partition_scheme_header_->getPartitionAttributeIds()) {
+        values.push_back(accessor->getTypedValue(attr_id));
+      }
```
hbdeshmukh: Did you take a look at the function ``getTupleWithAttributes()``. It may be useful here. Moreover the ``tuple`` class can compute hash using the ``getTupleHash`` function.

hbdeshmukh:
Hi @zuyu 

Finished one round of review. Please let me know when should I take another look.

catalog/PartitionSchemeHeader.hpp
zuyu: We use [lexicographical order](https://en.wikipedia.org/wiki/Lexicographical_order) here, `Given two different sequences of the same length, a1a2...ak and b1b2...bk, the first one is smaller than the second one for the lexicographical order, if ai<bi (for the order of A), for the first i where ai and bi differ.`

catalog/tests/PartitionScheme_unittest.cpp
zuyu: Sure, and we are talk about the same thing. In fact, `-2^(-1074)` is the largest negative floating point.

catalog/tests/PartitionScheme_unittest.cpp
zuyu: Yes, this test triggers [a debug check](https://travis-ci.org/apache/incubator-quickstep/jobs/213492259#L3473) in protobuf in the debug builds only.

storage/InsertDestination.cpp
zuyu: I just checked the methods you mentioned, but I don't think we could use them, as we also need to get partition id in a range partition.

query_optimizer/resolver/Resolver.cpp
zuyu: Both cases are covered in the parser, not in the optimizer.

catalog/PartitionSchemeHeader.cpp
zuyu: Done.

zuyu:
Hi @hbdeshmukh,

Thank you for the review. I have addressed most of issues, and please take another look.

And I also would like to propose to remove the debug builds in Travis CI.

catalog/tests/PartitionScheme_unittest.cpp
hbdeshmukh: Let's not check in commented code. We can either get rid of these tests or try to fix it. I am not sure what these tests are trying to verify. If the functionality being tested is not really critical, we can remove the commented code, create a JIRA issue and move on.

hbdeshmukh:
Hi @zuyu 

Looks good to me. I just have one last comment about the commented code in unit tests. After that, we should be good to go. Also, there are two commits for this PR, would you mind squashing them to one?

zuyu:
Hi @hbdeshmukh,

I have reported QUICKSTEP-86 for the test failure, and squashed all the commits. Please merge it. Thanks!

PR #224, opened 2017-03-27
Fix a bug in PackedPayloadHashTable

jianqiao:
This PR fixes a bug in `PackedPayloadHashTable` when `derived_accessor` is of type `TupleIdSequenceAdapterValueAccessor<ColumnVectorsValueAccesor>`.

Thanks Harshad @hbdeshmukh for pointing out the problem!

### DIFF
--- a/storage/PackedPayloadHashTable.cpp
+++ b/storage/PackedPayloadHashTable.cpp
@@ -235,30 +235,54 @@ bool PackedPayloadHashTable::upsertValueAccessorCompositeKey(
   ValueAccessor *base_accessor = accessor_mux.getBaseAccessor();
   ValueAccessor *derived_accessor = accessor_mux.getDerivedAccessor();
 
-  const bool has_derived_accessor = (derived_accessor != nullptr);
-
   base_accessor->beginIterationVirtual();
-  if (has_derived_accessor) {
+  if (derived_accessor != nullptr) {
+    // NOTE(jianqiao): Currently we expect derived_accessor to be either
+    // ColumnVectorsValueAccesor or
+    // TupleIdSequenceAdapterValueAccessor<ColumnVectorsValueAccesor>
     DCHECK(derived_accessor->getImplementationType()
                == ValueAccessor::Implementation::kColumnVectors);
+    DCHECK(!derived_accessor->isOrderedTupleIdSequenceAdapter());
     derived_accessor->beginIterationVirtual();
   }
 
   return InvokeOnBools(
-      has_derived_accessor,
       handles_.empty(),
       !all_keys_inline_,
-      [&](auto use_two_accessors,  // NOLINT(build/c++11)
-          auto key_only,
+      [&](auto key_only,  // NOLINT(build/c++11)
           auto has_variable_size) -> bool {
-    return this->upsertValueAccessorCompositeKeyInternal<
-        decltype(use_two_accessors)::value,
-        decltype(key_only)::value,
-        decltype(has_variable_size)::value>(
-            argument_ids,
-            key_attr_ids,
-            base_accessor,
-            static_cast<ColumnVectorsValueAccessor *>(derived_accessor));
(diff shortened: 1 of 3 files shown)


storage/PackedPayloadHashTable.cpp
```diff
@@ -239,26 +239,47 @@ bool PackedPayloadHashTable::upsertValueAccessorCompositeKey(
...
-        decltype(key_only)::value,
-        decltype(has_variable_size)::value>(
-            argument_ids,
-            key_attr_ids,
-            base_accessor,
-            static_cast<ColumnVectorsValueAccessor *>(derived_accessor));
+    constexpr bool key_only_v = decltype(key_only)::value;
+    constexpr bool has_variable_size_v = decltype(has_variable_size)::value;
+
+    if (!has_derived_accessor) {
```
zuyu: Minor suggestion regarding branch conditions: which is the most common case, `has_derived_accessor` is `true` or `false`?

storage/PackedPayloadHashTable.cpp
```diff
@@ -239,26 +239,47 @@ bool PackedPayloadHashTable::upsertValueAccessorCompositeKey(
...
+      return this->upsertValueAccessorCompositeKeyInternal<
+          false, key_only_v, has_variable_size_v>(
+              argument_ids,
+              key_attr_ids,
+              base_accessor,
+              static_cast<ColumnVectorsValueAccessor*>(nullptr));
+    } else if (derived_accessor->isTupleIdSequenceAdapter()) {
```
zuyu: Suggest to change to `if (derived_accessor->isTupleIdSequenceAdapter()) {`

storage/PackedPayloadHashTable.cpp
```diff
@@ -239,26 +239,47 @@ bool PackedPayloadHashTable::upsertValueAccessorCompositeKey(
...
-            argument_ids,
-            key_attr_ids,
-            base_accessor,
-            static_cast<ColumnVectorsValueAccessor *>(derived_accessor));
+    constexpr bool key_only_v = decltype(key_only)::value;
+    constexpr bool has_variable_size_v = decltype(has_variable_size)::value;
+
+    if (!has_derived_accessor) {
+      return this->upsertValueAccessorCompositeKeyInternal<
+          false, key_only_v, has_variable_size_v>(
```
zuyu: I'd suggest to change to `false /* has_derived_accessor */`, as well as in the following `true` cases.

storage/PackedPayloadHashTable.cpp
jianqiao: `false` is likely the common case. Just changed the condition to `derived_accessor == nullptr`.

storage/PackedPayloadHashTable.cpp
jianqiao: Updated.

jianqiao:
Updated to work around the gcc internal error.

hbdeshmukh:
Looks good to me. Additionally this branch improves TPC-H Q10's performance by nearly 50%. Will merge after travis succeeds.

zuyu:
Hi @hbdeshmukh, any intuition of what causes this huge performance boost? Thanks!

hbdeshmukh:
@zuyu There is a threshold to decide whether we perform the group by aggregation in a partitioned way or not. If the estimated number of groups (provided by the optimizer) is greater than the threshold, we go for partitioned aggregation. In TPC-H Q10 for SF100 the estimate comes out to be 50k and the default threshold is also 50k. Therefore we were not able to perform parallel partitioned aggregation. When I tried to force the aggregation to be partitioned, the addressed bug was exposed. With the changes in this PR, now the aggregation is partitioned and hence the performance improvement.

pateljm:
Nice job @hbdeshmukh and @jianqiao!

PR #226, opened 2017-04-02
Bug fix in memory measurement

hbdeshmukh:
- Some NULL checks added without which we may have segmentation faults.

### DIFF
--- a/storage/AggregationOperationState.cpp
+++ b/storage/AggregationOperationState.cpp
@@ -949,9 +949,15 @@ void AggregationOperationState::finalizeHashTableImplThreadPrivate(
 std::size_t AggregationOperationState::getMemoryConsumptionBytes() const {
   std::size_t memory = getMemoryConsumptionBytesHelper(distinctify_hashtables_);
   memory += getMemoryConsumptionBytesHelper(group_by_hashtables_);
-  memory += collision_free_hashtable_->getMemoryConsumptionBytes();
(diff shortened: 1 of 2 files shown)


storage/AggregationOperationState.cpp
```diff
@@ -949,9 +949,15 @@ void AggregationOperationState::finalizeHashTableImplThreadPrivate(
 std::size_t AggregationOperationState::getMemoryConsumptionBytes() const {
   std::size_t memory = getMemoryConsumptionBytesHelper(distinctify_hashtables_);
   memory += getMemoryConsumptionBytesHelper(group_by_hashtables_);
-  memory += collision_free_hashtable_->getMemoryConsumptionBytes();
-  memory += group_by_hashtable_pool_->getMemoryConsumptionPoolBytes();
-  memory += partitioned_group_by_hashtable_pool_->getMemoryConsumptionPoolBytes();
+  if (collision_free_hashtable_ != nullptr) {
```
cramja: Are these mutually exclusive? If so, could we use `else if` ?

hbdeshmukh: I think it is alright to have different if statements here.

jianqiao:
LGTM! Merging.

cramja:
+1

jianqiao:
Have merged the branch but not reflected in github.

PR #227, opened 2017-04-05
Added an enum type for all relational operators

hbdeshmukh:
The goal of this PR is that the scheduler should be able to identify individual operations in the query plan DAG.


hbdeshmukh:
Assigned to @jianqiao

jianqiao:
LGTM! Merging.

PR #230, opened 2017-04-08
Initial SingleNodeClient implementation

cramja:
This is the first part of a couple of PRs which will implement basic server functionality in QS. The basic usage of the client is:
```
$ GLOG_logtostderr=1 ./quickstep_client --port 3001 < my_query.sql

W0408 09:10:28.233783 3400758208 QuickstepClient.cpp:74] RPC call failed with code 14 and message: Connect Failed
RPC failed
```
The client uses gRPC libraries from the tmb and thus makes some changes to cli/CMakelists to include necessary libraries.

This will not add additional functionality on top of the the existing CLI, it merely allows quickstep to run as a separate process.

The next PR will have appropriate changes to the cli to handle network interactions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,6 +145,7 @@ if (ENABLE_VECTOR_PREDICATE_SHORT_CIRCUIT)
   )
 endif()
 
+option(ENABLE_SINGLE_NODE_SERVER "Enables server mode in single node setting" ON)
 option(ENABLE_DISTRIBUTED "Use the distributed version of Quickstep" OFF)
 
 macro (set_gflags_lib_name)
@@ -792,3 +793,13 @@ if (ENABLE_DISTRIBUTED)
                         ${GFLAGS_LIB_NAME}
                         ${GRPCPLUSPLUS_LIBRARIES})
 endif(ENABLE_DISTRIBUTED)
+
+if(ENABLE_SINGLE_NODE_SERVER)
+  add_executable (quickstep_client cli/QuickstepClient.cpp)
+  target_link_libraries(quickstep_client
+            ${GFLAGS_LIB_NAME}
+            ${GRPCPLUSPLUS_LIBRARIES}
+            ${PROTOBUF3_LIBRARIES}
+            glog
+            quickstep_cli_SingleNodeServer_proto)
+endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -795,11 +795,10 @@ if (ENABLE_DISTRIBUTED)
 endif(ENABLE_DISTRIBUTED)
 
 if(ENABLE_SINGLE_NODE_SERVER)
-  add_executable (quickstep_client cli/QuickstepClient.cpp)
+  add_executable (quickstep_client cli/SingleNodeClientExecutable.cpp)
   target_link_libraries(quickstep_client
             ${GFLAGS_LIB_NAME}
             ${GRPCPLUSPLUS_LIBRARIES}
-            ${PROTOBUF3_LIBRARIES}
             glog
-            quickstep_cli_SingleNodeServer_proto)
+            quickstep_cli_SingleNodeClient)
 endif()
--- a/cli/CMakeLists.txt
+++ b/cli/CMakeLists.txt
@@ -44,6 +44,22 @@ configure_file (
   "${CMAKE_CURRENT_BINARY_DIR}/CliConfig.h"
 )
 
+# Compile the protos for Single Node Server mode.
+if(ENABLE_SINGLE_NODE_SERVER)
+  # We will need some of the TMBs libraries
+  set(CMAKE_MODULE_PATH
+      ${CMAKE_MODULE_PATH}
+      "${PROJECT_SOURCE_DIR}/third_party/src/tmb/cmake")
+
+  find_package(Grpc++ REQUIRED)
+  include_directories(${GRPCPLUSPLUS_INCLUDE_DIRS})
+
+  GRPC_GENERATE_CPP(cli_SingleNodeServer_proto_srcs
+                    cli_SingleNodeServer_proto_hdrs
+                    .
+                    SingleNodeServer.proto)
+endif()
+
 # Declare micro-libs and link dependencies:
 add_library(quickstep_cli_CommandExecutor CommandExecutor.cpp CommandExecutor.hpp)
 add_library(quickstep_cli_CommandExecutorUtil CommandExecutorUtil.cpp CommandExecutorUtil.hpp)
@@ -65,6 +81,12 @@ else()
               LineReaderDumb.cpp
               LineReader.hpp
               LineReaderDumb.hpp)
+endif()
+
+if (ENABLE_SINGLE_NODE_SERVER)
+  add_library(quickstep_cli_SingleNodeServer_proto
+              ${cli_SingleNodeServer_proto_srcs}
+              ${cli_SingleNodeServer_proto_hdrs})
 endif()
 
 add_library(quickstep_cli_PrintToScreen PrintToScreen.cpp PrintToScreen.hpp)
@@ -150,7 +172,15 @@ if(USE_LINENOISE)
 else()
   target_link_libraries(quickstep_cli_LineReader
                         quickstep_utility_Macros)
+
 endif()
+
+if(ENABLE_SINGLE_NODE_SERVER)
+  target_link_libraries(quickstep_cli_SingleNodeServer_proto
+                        ${GRPCPLUSPLUS_LIBRARIES}
+                        ${PROTOBUF3_LIBRARY})
+endif()
+
 target_link_libraries(quickstep_cli_PrintToScreen
                       ${GFLAGS_LIB_NAME}
                       quickstep_catalog_CatalogAttribute
--- a/cli/CMakeLists.txt
+++ b/cli/CMakeLists.txt
@@ -84,6 +84,12 @@ else()
 endif()
 
 if (ENABLE_SINGLE_NODE_SERVER)
+  add_library(quickstep_cli_SingleNodeClient
+              SingleNodeClient.cpp
+              SingleNodeClient.hpp)
+  add_library(quickstep_cli_SingleNodeServerWrapper
+              SingleNodeServerWrapper.cpp
+              SingleNodeServerWrapper.hpp)
   add_library(quickstep_cli_SingleNodeServer_proto
               ${cli_SingleNodeServer_proto_srcs}
               ${cli_SingleNodeServer_proto_hdrs})
@@ -92,6 +98,14 @@ endif()
 add_library(quickstep_cli_PrintToScreen PrintToScreen.cpp PrintToScreen.hpp)
 
 # Link dependencies:
+if (ENABLE_SINGLE_NODE_SERVER)
+  target_link_libraries(quickstep_cli_SingleNodeClient
+                        ${GFLAGS_LIB_NAME}
+                        ${GRPCPLUSPLUS_LIBRARIES}
+                        ${PROTOBUF3_LIBRARIES}
+                        glog
+                        quickstep_cli_SingleNodeServer_proto)
+endif()
 target_link_libraries(quickstep_cli_CommandExecutor
                       glog
                       quickstep_catalog_CatalogAttribute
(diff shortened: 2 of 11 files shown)


cramja:
@zuyu is there something special we specify in Travis to get Proto3?

zuyu:
There are several ways to get `Protobuf 3` and `gRPC` in Travis, but unfortunately all may result in time-out.
 * PR #223 upgrades git submodule in the third_party.
 * PR #220 downloads `Protobuf 3` source code and builds locally.

cramja:
Oh shoot, so no Travis for dist?

zuyu:
Unfortunately, for now no CIs for the distributed version due to both `Protobuf 3` and `gRPC` dependencies. But we definitely should have!

One thing to add is that ccache could speed up the build time to avoid time-out, but I have no luck to set it up correctly.

CMakeLists.txt
```diff
@@ -696,7 +697,8 @@ set(ENABLE_ZOOKEEPER OFF CACHE BOOL "Enable Zookeeper TMB")
 add_subdirectory("${THIRD_PARTY_SOURCE_DIR}/tmb" "${CMAKE_CURRENT_BINARY_DIR}/third_party/tmb")
 include_directories(${TMB_INCLUDE_DIRS})
 
-if (ENABLE_DISTRIBUTED)
+# Single node server requires this for gRPC headers
+if (ENABLE_DISTRIBUTED OR ENABLE_SINGLE_NODE_SERVER)
```
zuyu: Reset the change, as `gRPC` does not come from `TMB`.

CMakeLists.txt
```diff
@@ -792,3 +794,13 @@ if (ENABLE_DISTRIBUTED)
...
 endif(ENABLE_DISTRIBUTED)
+
+if(ENABLE_SINGLE_NODE_SERVER)
+  add_executable (quickstep_client cli/QuickstepClient.cpp)
+  target_link_libraries(quickstep_client
+            ${GFLAGS_LIB_NAME}
+            glog
+            ${GRPCPLUSPLUS_LIBRARIES}
+            ${PROTOBUF3_LIBRARIES}
+            quickstep_cli_SingleNodeServer_proto)
```
zuyu: All linking libraries should be aligned with `(`, and sort in the alphabet order.

```
             glog
             quickstep_cli_SingleNodeServer_proto
             ${GFLAGS_LIB_NAME}
             ${GRPCPLUSPLUS_LIBRARIES})
```

cli/CMakeLists.txt
```diff
@@ -44,6 +44,20 @@ configure_file (
   "${CMAKE_CURRENT_BINARY_DIR}/CliConfig.h"
 )
 
+# Compile the protos for Single Node Server mode.
+if(ENABLE_SINGLE_NODE_SERVER)
+  # We will need some of the TMBs libraries
+  set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/third_party/src/tmb/cmake")
+
+  find_package(Grpc++ REQUIRED)
+  include_directories(${GRPCPLUSPLUS_INCLUDE_DIRS})
```
zuyu: We should move above to the root `cmake` file. Also, it should look like

```
  set(CMAKE_MODULE_PATH "$CMAKE_MODULE_PATH"
                        "${PROJECT_SOURCE_DIR}/third_party/src/tmb/cmake")

  find_package(Grpc++ REQUIRED)
  include_directories(${GRPCPLUSPLUS_INCLUDE_DIRS})
```

cli/CMakeLists.txt
```diff
@@ -150,7 +170,15 @@ if(USE_LINENOISE)
 else()
   target_link_libraries(quickstep_cli_LineReader
                         quickstep_utility_Macros)
+
 endif()
+
+if(ENABLE_SINGLE_NODE_SERVER)
+  target_link_libraries(quickstep_cli_SingleNodeServer_proto
+                        ${PROTOBUF_LIBRARY}
```
zuyu: Should be `${PROTOBUF3_LIBRARIES}`.

cli/QuickstepClient.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "cli/SingleNodeServer.grpc.pb.h"
+#include "cli/SingleNodeServer.pb.h"
+
+#include "gflags/gflags.h"
+#include "glog/logging.h"
+#include "grpc++/grpc++.h"
```
zuyu: Should be `<grpc++/grpc++.h>`.

cli/QuickstepClient.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+#include <chrono>
+#include <cstddef>
+#include <cstdio>
+#include <exception>
+#include <iostream>
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
```
zuyu: Please run `iwyu` for all new source code files.

cli/QuickstepClient.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include "gflags/gflags.h"
+#include "glog/logging.h"
+#include "grpc++/grpc++.h"
+
+using grpc::Channel;
+using grpc::ClientContext;
+using grpc::Status;
+
+static bool ValidatePort(const char *flagname, std::int32_t value) {
+  if (value >= 0 && value < 32768)
```
zuyu: A valid value should be `(0, 65536)`, not including both boundaries.

cli/QuickstepClient.cpp
```diff
@@ -0,0 +1,109 @@
...
+    return true;
+  printf("Invalid value for --%s: %d\n", flagname, (int) value);
```
zuyu: Use `std::printf`.

cli/QuickstepClient.cpp
```diff
@@ -0,0 +1,109 @@
...
+  return false;
+}
+
+DEFINE_int32(port, 3000, "Listens for TCP connection on this port.");
+DEFINE_validator(port, &ValidatePort);
+
+namespace quickstep {
+
+class QuickstepClient {
+public:
```
zuyu: Add one whitespace indentation. Similarly for `private:` below.

cli/QuickstepClient.cpp
```diff
@@ -0,0 +1,109 @@
...
+  QuickstepClient(std::shared_ptr<Channel> channel)
+    : stub_(SingleNodeServerRequest::NewStub(channel)) {}
```
zuyu: Add two more whitespace indentation.

cli/QuickstepClient.cpp
```diff
@@ -0,0 +1,109 @@
...
+}
+
+DEFINE_int32(port, 3000, "Listens for TCP connection on this port.");
+DEFINE_validator(port, &ValidatePort);
+
+namespace quickstep {
+
+class QuickstepClient {
+public:
+  QuickstepClient(std::shared_ptr<Channel> channel)
```
zuyu: Use `const reference` to avoid a deep copy of a smart pointer.

cli/QuickstepClient.cpp
```diff
@@ -0,0 +1,109 @@
...
+    }
+  }
+
+private:
+  /**
+   * Handle a valid response from the server.
+   * @param response A valid query response.
+   * @return The response string.
+   */
+  std::string HandleQueryResponse(QueryResponse const &response) {
```
zuyu: Use `std::string HandleQueryResponse(const QueryResponse & response) const {`.

In fact, it could be a static, non-member method..

cli/QuickstepClient.cpp
```diff
@@ -0,0 +1,109 @@
...
+    if(response.error()) {
+      return response.error_result();
+    }
+    return response.query_result();
+  }
+
+  std::unique_ptr<SingleNodeServerRequest::Stub> stub_;
+};
+}
```
zuyu: Code style: add an empty line and change to `}  // quickstep`.

cli/QuickstepClient.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+int main(int argc, char** argv) {
+  google::InitGoogleLogging(argv[0]);
+  gflags::ParseCommandLineFlags(&argc, &argv, true);
+
+  // Attempts to send a single query retrieved from stdin to the Quickstep Server.
+  quickstep::QuickstepClient qs_client(
+    grpc::CreateChannel("localhost:" + std::to_string(FLAGS_port),
+    grpc::InsecureChannelCredentials()));
```
zuyu: Both lines above should add two more whitespace indentations.

cli/QuickstepClient.cpp
```diff
@@ -0,0 +1,109 @@
...
+  std::string user_query;
+  std::cin >> user_query;
```
zuyu: We really need some tests. This does not work, even for a simply `SELECT-one` query, as it can not handle an input with whitespace.

cli/SingleNodeServer.proto
```diff
@@ -0,0 +1,34 @@
...
+
+message QueryRequest {
+    string query = 1;
+}
+
+message QueryResponse {
+    bool error = 1;
+    string query_result = 2;
+    string error_result = 3;
+}
```
zuyu: Please reduce to two whitespace indentations, and add `EOL`.

cli/SingleNodeServer.proto
```diff
@@ -0,0 +1,34 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+syntax = "proto3";
+
+package quickstep;
+
+service SingleNodeServerRequest {
```
zuyu: Recommend to rename to `SingleNodeServer`, so that the service name is the same as the file name.

cramja:
@zuyu Thank you for the review

Here's a possibility based on your overall comment: I could add the IOBase and its subclass GrpcWrapper to this PR. The IOBase abstracts around all IO (including LineReaders). The GrpcWrapper is a wrapper around server side interactions. Then the unit test would test communication between GrpcWrapper and the Client.

I was thinking to add 1 new executable to Quickstep, that being the dumb client. All it really is is an interface for a user to send a query through. It just handles send and receive, therefore does no parsing/etc. That can come later if we want.

Then in another PR the Cli exe can be modified to have the `--mode=server` option as you suggested.

cli/CMakeLists.txt
cramja: Seems to me like leaving it in `/cli/cmakelists.txt` should be fine because the module is being used by sources in `/cli/` but not necessarily by sources in `/`.

cramja: Also good catch on keeping the orig module path

Note that cmake likes brackets:
```
set(CMAKE_MODULE_PATH
      ${CMAKE_MODULE_PATH}
      "${PROJECT_SOURCE_DIR}/third_party/src/tmb/cmake")
```

works while

```
set(CMAKE_MODULE_PATH
      $CMAKE_MODULE_PATH
      "${PROJECT_SOURCE_DIR}/third_party/src/tmb/cmake")
```
treats the middle argument like a string literal

cramja: Also not that because of CMAKE scoping, the module will not be added to the parent scope's module path:

`/cmakelists.txt`:
```
message("before module path: ${CMAKE_MODULE_PATH}")
add_subdirectory(cli)
message("after module path: ${CMAKE_MODULE_PATH}")
```
output:
```
before module path: /Users/cramja/workspace/quickstep/incubator-quickstep/cmake
after module path: /Users/cramja/workspace/quickstep/incubator-quickstep/cmake
```

CMakeLists.txt
```diff
@@ -145,6 +145,7 @@ if (ENABLE_VECTOR_PREDICATE_SHORT_CIRCUIT)
   )
 endif()
 
+option(ENABLE_SINGLE_NODE_SERVER "Enables server mode in single node setting" ON)
```
cramja: Does it need to be for any reason other than travis?

cli/QuickstepClient.cpp
cramja: I removed some of the obviously not needed includes. However, these tools are imprecise, and I find that their recommendations don't always work, or agree with one another. I.e. Clion IDE gives me recommendations as well but they are different from iwyu.

cli/SingleNodeServer.proto
cramja: maybe `SingleNodeServerService` ?

CMakeLists.txt
zuyu: Yes, for now `Quickstep` does not declare the `grpc` dependency. So, by default, we don't build any binaries that link to `grpc`.

cli/SingleNodeServer.proto
zuyu: Then, change the file name as well.

zuyu:
For this PR, we could add two new classes: `client` and `server`, and a unit test to test the communication path between both. No need changes to any existing source code files.

cramja:
@zuyu I'm going to close this PR and submit another one with the tests as I think the content has changed sufficiently enough to have another PR

PR #232, opened 2017-04-17
QUICKSTEP-87 Adds network cli interface.

cramja:
# Adds Network CLI functionality to Single Node

**Functional Changes include:**
]Adding a network mode to the cli. Run quickstep in `network` mode (defaults to `local`)

`quickstep_cli_shell --mode=network`

Now quickstep boots a server, you can interact using the client executable...

`quickstep_client < my_query.sql`

**Code Changes include:**
* Adds a generic IO interface
  * Hides Linereader from CLI behind class `LocalIO`
  * New `NetworkIO` also hidden here
* Adds tests for NetworkIO
* Adds simple Client executable 
* Cmake flag to disable (`ENABLE_NETWORK_CLI=false`)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,6 +145,7 @@ if (ENABLE_VECTOR_PREDICATE_SHORT_CIRCUIT)
   )
 endif()
 
+option(ENABLE_NETWORK_CLI "Allows use of the network cli" OFF)
 option(ENABLE_DISTRIBUTED "Use the distributed version of Quickstep" OFF)
 
 macro (set_gflags_lib_name)
@@ -730,6 +731,7 @@ target_link_libraries(quickstep_cli_shell
                       quickstep_cli_Flags
                       quickstep_cli_InputParserUtil
                       quickstep_cli_LineReader
+                      quickstep_cli_LocalIO
                       quickstep_cli_PrintToScreen
                       quickstep_parser_ParseStatement
                       quickstep_parser_SqlParserWrapper
@@ -755,7 +757,10 @@ if (ENABLE_HDFS)
   target_link_libraries(quickstep_cli_shell
                         quickstep_storage_FileManagerHdfs)
 endif()
-
+if (ENABLE_NETWORK_CLI)
+  target_link_libraries(quickstep_cli_shell
+                        quickstep_cli_NetworkIO)
+endif()
 # Link against other required system and third-party libraries.
 target_link_libraries(quickstep_cli_shell ${LIBS})
 
@@ -792,3 +797,14 @@ if (ENABLE_DISTRIBUTED)
                         ${GFLAGS_LIB_NAME}
                         ${GRPCPLUSPLUS_LIBRARIES})
 endif(ENABLE_DISTRIBUTED)
+
+if (ENABLE_NETWORK_CLI)
+  add_executable (quickstep_client cli/NetworkCliClientMain.cpp)
+  target_link_libraries(quickstep_client
+                        ${GFLAGS_LIB_NAME}
+                        ${GRPCPLUSPLUS_LIBRARIES}
+                        glog
+                        quickstep_cli_LineReaderBuffered
+                        quickstep_cli_NetworkCliClient
+                        quickstep_cli_NetworkIO)
+endif()
--- a/cli/CMakeLists.txt
+++ b/cli/CMakeLists.txt
@@ -16,6 +16,7 @@
 # under the License.
 
 include_directories(${CMAKE_CURRENT_BINARY_DIR})
+
 if (ENABLE_DISTRIBUTED)
   add_subdirectory(distributed)
 endif(ENABLE_DISTRIBUTED)
@@ -39,10 +40,30 @@ if (ENABLE_GOOGLE_PROFILER)
   set(QUICKSTEP_ENABLE_GOOGLE_PROFILER TRUE)
 endif()
 
+if (ENABLE_NETWORK_CLI)
+  set(QUICKSTEP_ENABLE_NETWORK_CLI TRUE)
+endif()
+
 configure_file (
   "${CMAKE_CURRENT_SOURCE_DIR}/CliConfig.h.in"
   "${CMAKE_CURRENT_BINARY_DIR}/CliConfig.h"
 )
+
+# Compile the protos for Single Node Server mode.
+if (ENABLE_NETWORK_CLI)
+  # We will need some of the TMBs libraries
+  set(CMAKE_MODULE_PATH
+      ${CMAKE_MODULE_PATH}
+      "${PROJECT_SOURCE_DIR}/third_party/src/tmb/cmake")
+
+  find_package(Grpc++ REQUIRED)
+  include_directories(${GRPCPLUSPLUS_INCLUDE_DIRS})
+
+  GRPC_GENERATE_CPP(cli_NetworkCli_proto_srcs
+                    cli_NetworkCli_proto_hdrs
+                    .
+                    NetworkCli.proto)
+endif()
 
 # Declare micro-libs and link dependencies:
 add_library(quickstep_cli_CommandExecutor CommandExecutor.cpp CommandExecutor.hpp)
@@ -51,6 +72,7 @@ add_library(quickstep_cli_Constants ../empty_src.cpp Constants.hpp)
 add_library(quickstep_cli_DefaultsConfigurator DefaultsConfigurator.cpp DefaultsConfigurator.hpp)
 add_library(quickstep_cli_DropRelation DropRelation.cpp DropRelation.hpp)
 add_library(quickstep_cli_Flags Flags.cpp Flags.hpp)
+add_library(quickstep_cli_IOInterface ../empty_src.cpp IOInterface.hpp)
 add_library(quickstep_cli_InputParserUtil InputParserUtil.cpp InputParserUtil.hpp)
 
 if(USE_LINENOISE)
@@ -65,6 +87,25 @@ else()
               LineReaderDumb.cpp
               LineReader.hpp
               LineReaderDumb.hpp)
+endif()
+
+add_library(quickstep_cli_LineReaderBuffered
+            LineReaderBuffered.cpp
+            LineReaderBuffered.hpp)
+add_library(quickstep_cli_LocalIO
+            ../empty_src.cpp
+            LocalIO.hpp)
+
+if (ENABLE_NETWORK_CLI)
+  add_library(quickstep_cli_NetworkCli_proto
+              ${cli_NetworkCli_proto_srcs}
+              ${cli_NetworkCli_proto_hdrs})
+  add_library(quickstep_cli_NetworkCliClient
+              ../empty_src.cpp
+              NetworkCliClient.hpp)
+  add_library(quickstep_cli_NetworkIO
+              NetworkIO.cpp
+              NetworkIO.hpp)
 endif()
 
 add_library(quickstep_cli_PrintToScreen PrintToScreen.cpp PrintToScreen.hpp)
@@ -135,6 +176,8 @@ target_link_libraries(quickstep_cli_Flags
                       quickstep_cli_DefaultsConfigurator
                       quickstep_storage_StorageConstants
                       ${GFLAGS_LIB_NAME})
+target_link_libraries(quickstep_cli_IOInterface
+                      quickstep_utility_Macros)
 target_link_libraries(quickstep_cli_InputParserUtil
                       glog
                       quickstep_utility_Macros
@@ -150,6 +193,37 @@ if(USE_LINENOISE)
 else()
   target_link_libraries(quickstep_cli_LineReader
                         quickstep_utility_Macros)
+endif()
+target_link_libraries(quickstep_cli_LineReaderBuffered
+                      quickstep_cli_LineReader
+                      quickstep_utility_Macros)
+target_link_libraries(quickstep_cli_LocalIO
+                      quickstep_cli_LineReader
+                      quickstep_cli_IOInterface
+                      quickstep_utility_Macros)
+if (ENABLE_NETWORK_CLI)
+  target_link_libraries(quickstep_cli_NetworkCli_proto
+                        ${GRPCPLUSPLUS_LIBRARIES}
+                        ${PROTOBUF3_LIBRARY})
+  target_link_libraries(quickstep_cli_NetworkCliClient
+                        ${GFLAGS_LIB_NAME}
+                        ${GRPCPLUSPLUS_LIBRARIES}
+                        ${PROTOBUF3_LIBRARIES}
+                        glog
+                        quickstep_cli_NetworkCli_proto
+                        quickstep_utility_Macros)
+  target_link_libraries(quickstep_cli_NetworkIO
+                        ${GFLAGS_LIB_NAME}
+                        ${GRPCPLUSPLUS_LIBRARIES}
+                        ${PROTOBUF3_LIBRARIES}
+                        glog
+                        quickstep_cli_IOInterface
+                        quickstep_cli_NetworkCli_proto
+                        quickstep_threading_ConditionVariable
+                        quickstep_threading_SpinSharedMutex
+                        quickstep_utility_Macros
+                        quickstep_utility_MemStream
+                        quickstep_utility_ThreadSafeQueue)
 endif()
 target_link_libraries(quickstep_cli_PrintToScreen
                       ${GFLAGS_LIB_NAME}
@@ -175,6 +249,15 @@ target_link_libraries(quickstep_cli
                       quickstep_cli_DefaultsConfigurator
                       quickstep_cli_DropRelation
                       quickstep_cli_Flags
+                      quickstep_cli_IOInterface
                       quickstep_cli_InputParserUtil
                       quickstep_cli_LineReader
+                      quickstep_cli_LineReaderBuffered
+                      quickstep_cli_LocalIO
                       quickstep_cli_PrintToScreen)
+if (ENABLE_NETWORK_CLI)
+  target_link_libraries(quickstep_cli
+                        quickstep_cli_NetworkCli_proto
+                        quickstep_cli_NetworkCliClient
+                        quickstep_cli_NetworkIO)
+endif()
--- a/cli/IOInterface.hpp
+++ b/cli/IOInterface.hpp
@@ -1,0 +1,84 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ **/
+
+#ifndef QUICKSTEP_CLI_IO_INTERFACE_HPP_
+#define QUICKSTEP_CLI_IO_INTERFACE_HPP_
+
+#include <cstdio>
+#include <string>
+
+#include "utility/Macros.hpp"
+
+namespace quickstep {
+
+/**
+ * An individual IO interaction with Quickstep.
+ */
+class IOHandle {
+ public:
+  IOHandle() {}
+
+  /**
+   * @note Destructor should handle clean up of any IO state.
+   */
+  virtual ~IOHandle() {}
+
+  /**
+   * @return A file handle for standard output.
+   */
+  virtual FILE *out() = 0;
+
+  /**
+   * @return A file handle for error output.
+   */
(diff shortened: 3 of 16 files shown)


hbdeshmukh:
Hi @cramja 

Considering that this is a fairly big feature, can you please create a JIRA ticket and link this PR with it?

cramja:
@hbdeshmukh I created a ticket `QUICKSTEP-87`; maybe you can show me how to link to this PR later.

cramja:
Thanks @hbdeshmukh , fixed validate_cmakelists issue.

cli/Flags.cpp
```diff
@@ -87,4 +87,15 @@ DEFINE_bool(preload_buffer_pool, false,
             "accepting queries (should also set --buffer_pool_slots to be "
             "large enough to accomodate the entire database).");
 
+static bool ValidatePort(const char *flagname, std::int32_t value) {
+  if (value > 0 && value < 65536) {
+    return true;
+  }
+  std::cout << "Invalid value for --" << flagname << ": " << std::to_string(value) << std::endl;
```
hbdeshmukh: Should the error message also include the valid range of values? That might be helpful.

cli/IOInterface.hpp
```diff
@@ -0,0 +1,61 @@
...
+   */
+  virtual FILE* out() = 0;
+
+  /**
+   * @return A file handle for error output.
+   */
+  virtual FILE* err() = 0;
+
+  /**
+   * Requests a complete SQL command. This call may block until user input is given.
```
hbdeshmukh: Missing the @brief annotation

cli/IOInterface.hpp
```diff
@@ -0,0 +1,61 @@
...
+   * @note When the command is complete, commandComplete() should be called so that certain implementations can clean
+   *    up state related to sending the command.
+   * @return A string containing a quickstep command.
+   */
+  virtual std::string nextCommand() = 0;
+
+  /**
+   * Notifies the IO system that the previously acquired command is complete.
```
hbdeshmukh: Missing the @ brief annotation

cli/IOInterface.hpp
```diff
@@ -0,0 +1,61 @@
...
+   */
+  virtual void commandComplete() {}
+
+  /**
+   * Notifies the IO system that quickstep is shutting down.
```
hbdeshmukh: Missing the @ brief annotation

cli/IOInterface.hpp
```diff
@@ -0,0 +1,61 @@
...
+   */
+  virtual FILE* err() = 0;
+
+  /**
+   * Requests a complete SQL command. This call may block until user input is given.
+   * @note When the command is complete, commandComplete() should be called so that certain implementations can clean
+   *    up state related to sending the command.
+   * @return A string containing a quickstep command.
+   */
+  virtual std::string nextCommand() = 0;
```
hbdeshmukh: Would ``getNextCommand`` be more intuitive?

cli/IOInterface.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+   * Notifies the IO system that the previously acquired command is complete.
+   */
+  virtual void commandComplete() {}
```
hbdeshmukh: How about ``notifyCommandComplete`` as a function name? In general, I try to name functions as a verb, so that their name can express what they are doing.

cli/IOInterface.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+   * Notifies the IO system that quickstep is shutting down.
+   */
+  virtual void shutdown() {}
```
hbdeshmukh: Similar to earlier, how about ``notifyShutdown``?

cli/LineReaderBuffered.cpp
```diff
@@ -0,0 +1,67 @@
...
+    if (command != "") {
+      stmts_vec.push_back(command);
+    }
+  }
+
+  EXPECT_EQ(stmt_1, stmts_vec.front());
+  EXPECT_EQ(4, stmts_vec.size());
+  EXPECT_TRUE(linereader.bufferEmpty());
+}
+*/
```
hbdeshmukh: Can you either uncomment the dead code or remove it?

cli/LineReaderBuffered.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+namespace quickstep {
+
+class LineReaderBuffered : public LineReader {
+ public:
+  /**
+   * A line reader which accepts a string buffer.
+   * This does no IO and ignores any prompts given to the constructor.
+   * @param default_prompt Ignored.
+   * @param continue_prompt Ignored.
```
hbdeshmukh: I don't quite understand the doxygen comments for the constructor arguments. Can you please elaborate?

cli/LineReaderBuffered.hpp
```diff
@@ -0,0 +1,70 @@
...
+ **/
+
+#ifndef QUICKSTEP_CLI_LINE_READER_BUFFERED_HPP_
+#define QUICKSTEP_CLI_LINE_READER_BUFFERED_HPP_
+
+#include "cli/LineReader.hpp"
+
+namespace quickstep {
+
+class LineReaderBuffered : public LineReader {
```
hbdeshmukh: Does ``BufferedLineReader`` sound a better name?

cli/LineReaderBuffered.cpp
cramja: Ah, I was stashing this the old fashion way... removed

cramja: ..well, actually I moved the commented out comments back into the unit test. Originally I had written this LRBuffered class, then removed it, placing the files somewhere safe. When I moved them back in, I forgot that I stashed the test in the file.

cli/LineReaderBuffered.hpp
cramja: I agree, though I was trying to keep consistency with the 2 other LR's: LRLineNoise, LRDumb

cli/LineReaderBuffered.hpp
cramja: how about:
```
  /**
   * @brief A line reader which accepts a string buffer.
   * Other line readers are ment to support some form of user interaction. This linereader does not and is intended for
   * programmer use- it does not print anything to stdout. Therefore it ignores any prompt strings given to the
   * inherited constructor.
   * @param default_prompt Not used by this line reader, but required by interface.
   * @param continue_prompt Not used by this line reader, but required by interface.
   */
```

cli/NetworkIO.hpp
```diff
@@ -0,0 +1,227 @@
...
+};
+
+class NetworkIO : public IOInterface {
+ public:
+  NetworkIO()
+      : IOInterface(),
+        out_stream_(),
+        err_stream_(),
+        server_(nullptr),
+        service_() {
```
zuyu: No needs for explicitly calling default constructors.

cli/NetworkIO.hpp
```diff
@@ -0,0 +1,227 @@
...
+    return out_stream_.file();
+  }
+
+  FILE* err() override {
+    return err_stream_.file();
+  }
+
+  std::string getNextCommand() override {
+    // critical section: wait for a command
+    networkio_internal::RequestState& request_state = service_.getRequestState();
```
zuyu: Change to `networkio_internal::RequestState &request_state`.

Ditto in `notifyCommandComplete`.

cli/QuickstepCli.cpp
```diff
@@ -28,26 +28,23 @@
 #include <utility>
 #include <vector>
 
-#include "cli/CliConfig.h"  // For QUICKSTEP_USE_LINENOISE, QUICKSTEP_ENABLE_GOOGLE_PROFILER.
+#include "cli/CliConfig.h"  // For QUICKSTEP_ENABLE_GOOGLE_PROFILER.
```
zuyu: Add `QUICKSTEP_ENABLE_NETWORK_CLI`.

cli/tests/CMakeLists.txt
```diff
@@ -90,3 +101,18 @@ if (ENABLE_DISTRIBUTED)
                         ${GFLAGS_LIB_NAME}
                         ${LIBS})
 endif(ENABLE_DISTRIBUTED)
+if (ENABLE_SINGLE_NODE_SERVER)
```
zuyu: Please add an empty line above.

cli/tests/NetworkIO_unittest.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+/**
+ * Contains a server instance for testing.
+ */
+class TestNetworkIO {
+ public:
+  TestNetworkIO()
+      : service_(),
+        server_address_("localhost:" + std::to_string(FLAGS_port)),
+        server_(nullptr) {
```
zuyu: No needs for default constructors, including `service_` and `server_`.

cli/tests/NetworkIO_unittest.cpp
```diff
@@ -0,0 +1,182 @@
...
+    builder.AddListeningPort(server_address_, grpc::InsecureServerCredentials());
+    builder.RegisterService(&service_);
+    server_ = builder.BuildAndStart();
+    CHECK(server_) << "Failed to start server";
+    LOG(INFO) << "TestSingleNodeServer listening on " << server_address_;
+  }
+
+  // Gets a message from the Service worker.
+  std::string getSentMessage() {
+    networkio_internal::RequestState& requestState = service_.getRequestState();
```
zuyu: Change to `networkio_internal::RequestState &request_state`.

Ditto in `setResponse`.

cli/tests/NetworkIO_unittest.cpp
```diff
@@ -0,0 +1,182 @@
...
+/**
+ * Tests that killing the service will cancel requests.
+ */
+TEST(NetworkIOTest, TestShutdown) {
+  TestNetworkIO server;
+
+  server.getService().kill();
+
+  NetworkCliClient client(
+    grpc::CreateChannel("localhost:" + std::to_string(FLAGS_port),
```
zuyu: Add two more whitespace indentations.

cli/tests/NetworkIO_unittest.cpp
```diff
@@ -0,0 +1,182 @@
...
+  Status status = client.SendQuery(request, &response);
+  ASSERT_EQ(status.error_code(), grpc::CANCELLED);
+}
+
+/**
+ * Tests a simple call and response to the Service.
+ */
+TEST(NetworkIOTest, TestNetworkIOCommandInteraction) {
+  NetworkIO networkIO;
+  std::string const query_stmt = kQueryRequest + ";" + kQueryRequest;
```
zuyu: Do you mean `const std::string query_statement`?

Also, why the second query is not ended with `;`?

cli/tests/NetworkIO_unittest.cpp
```diff
@@ -0,0 +1,182 @@
...
+    std::string command = networkIO.getNextCommand();
+    EXPECT_EQ(command, query_stmt);
+
+    // Set some output for the main test thread, return.
+    fprintf(networkIO.out(), "%s", kQueryResponse.c_str());
+    networkIO.notifyCommandComplete();
+  });
+
+  NetworkCliClient client(
+    grpc::CreateChannel("localhost:" + std::to_string(FLAGS_port),
```
zuyu: Add two more whitespace indentations.

cli/tests/NetworkIO_unittest.cpp
```diff
@@ -0,0 +1,182 @@
...
+                        grpc::InsecureChannelCredentials()));
+  QueryRequest request;
+  request.set_query(query_stmt);
+  QueryResponse response;
+  Status status = client.SendQuery(request, &response);
+  ASSERT_TRUE(status.ok());
+  EXPECT_EQ(response.query_result(), kQueryResponse);
+  EXPECT_EQ(response.error_result(), std::string(""));
```
zuyu: Change to

```
EXPECT_EQ(kQueryResponse, response.query_result());
EXPECT_TRUE(response.error_result().empty());
```

cli/tests/NetworkIO_unittest.cpp
```diff
@@ -0,0 +1,182 @@
...
+  server_handler.join();
+}
+
+TEST(NetworkIOTest, TestLineReaderBuffered) {
+  // The buffered line reader is used exclusively by the NetworkIO's client.
+  LineReaderBuffered linereader;
+  EXPECT_TRUE(linereader.bufferEmpty());
+
+  std::string stmt = "select * from foo;";
+  std::string stmts = stmt + "select 1; select 2; quit;";
```
zuyu: Suggest to change to `single_statement` and `multiple_statements`.

CMakeLists.txt
```diff
@@ -145,6 +145,7 @@ if (ENABLE_VECTOR_PREDICATE_SHORT_CIRCUIT)
   )
 endif()
 
+option(ENABLE_NETWORK_CLI "Allows use of the network cli" ON)
```
cramja: Yes, I remembered. However, this was only because of Travis and I'd therefore prefer to keep this available by default. So I disabled it in the `.travis.yml` file.

cramja:
@zuyu Thank you for review. Regarding the question about mutexes and cond_var in the `NetworkCliServiceImpl`, we disallow multiple gRPC threads from submitting requests at the same time.

As for the suggestion, I think it's okay to split those up. I didn't do this in the draft because it creates very small test files.

cramja:
I should also mention that @hbdeshmukh and I had a discussion yesterday about refactoring the CLI main method. I was unhappy with how cluttered seeming it has become, and this PR only makes it slightly more cluttered. I came up with a solution which I have been working in a WIP branch. It uses RIIA to close down any IO-related state (eg network connections) by creating an `IOHandle` for every cli user interaction. Because I think this change is a big improvement over, and complimentary to the change which is present here, I will submit it as a commit on top of this commit in the PR.

cli/NetworkCliClient.hpp
```diff
@@ -0,0 +1,91 @@
...
+using grpc::Status;
+
+namespace quickstep {
+
+/**
+ * A simple wrapper class used to do CLI interactions with QuickstepCLI via the gRPC interface.
+ */
+class NetworkCliClient {
+ public:
+  explicit NetworkCliClient(std::shared_ptr<Channel> const &channel)
```
cramja: FYI the style guide says this is a matter of preference

> That said, while we encourage putting const first, we do not require it. But be consistent with the code around you!

My opinion is that it should always follow the target object because then it will always be consistent.

cramja: That said, I will change it here because elsewhere in this file, the const-first pattern is followed.

cli/tests/CMakeLists.txt
```diff
@@ -32,6 +32,17 @@ if (ENABLE_DISTRIBUTED)
                  "${PROJECT_SOURCE_DIR}/utility/textbased_test/TextBasedTest.hpp")
 endif(ENABLE_DISTRIBUTED)
 
+if (ENABLE_SINGLE_NODE_SERVER)
```
cramja: Corrected to `ENABLE_NETWORK_CLI`

cli/NetworkCliClient.hpp
zuyu: By consistency, we should be aligned with all usages in the codebase. Thus, `const Foo &foo`.

CMakeLists.txt
zuyu: The problem is, if we check in this PR, by default we will build the network-based client for the single-node version, no matter whether the machine has `grpc` installed or not. And typically, it will fail due to no `grpc` found.

cli/NetworkIO.hpp
```diff
@@ -0,0 +1,286 @@
...
+    // Go through each pending request, and cancel them.
+    while (!request_queue_.empty()) {
+      request_queue_.popOne()->cancel();
+    }
+  }
+
+ private:
+  Mutex service_mtx_;
+  bool running_;
+  ThreadSafeQueue<RequestState*> request_queue_;
```
zuyu: I suggest to follow [the official async server example](https://github.com/grpc/grpc/blob/master/examples/cpp/helloworld/greeter_async_server.cc#L54) so that we don't need mutexes and locks for multiple concurrent requests.

cli/Flags.cpp
```diff
@@ -87,4 +87,25 @@ DEFINE_bool(preload_buffer_pool, false,
...
+            << max << std::endl;
+  return false;
+}
+
+DEFINE_int32(cli_network_port, 3000,
+             "Listens for TCP connections on this port when network mode is enabled. "
+             "This is only used if the cli is set to use the network mode (--mode=network).");
+DEFINE_validator(cli_network_port, &ValidatePort);
+
+DEFINE_string(cli_network_ip, "localhost",
```
cramja: `0.0.0.0` is good. I originally tried `192.0.0.1` but that caused grpc to fail.

cramja: Also the CliClientMain uses the flag so I think they are appropriate to be kept there.

cli/CMakeLists.txt
```diff
@@ -15,7 +15,6 @@
 # specific language governing permissions and limitations
 # under the License.
 
-include_directories(${CMAKE_CURRENT_BINARY_DIR})
```
cramja: As I was checking over the commit, I saw this and thought I added it so that the compiled grpc files would be included. On looking at the history, it was not my addition- I'll put it back.

But it raises the question: why is it there to begin with?

cli/IOInterface.hpp
```diff
@@ -0,0 +1,82 @@
...
+   * @return A file handle for standard output.
+   */
+  virtual FILE *out() = 0;
+
+  /**
+   * @return A file handle for error output.
+   */
+  virtual FILE *err() = 0;
+
+  virtual std::string getCommand() = 0;
```
cramja: Calling `getCommand` could cause the interface to change state (pop an IOHandle off a queue, for example)

cramja: Err, I was thinking of `virtual IOHandle* getNextIOHandle()`. Your change makes sense.

cli/NetworkCliClientMain.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+#include "gflags/gflags.h"
+
+using quickstep::LineReaderBuffered;
+using quickstep::NetworkCliClient;
+
+int main(int argc, char **argv) {
+  google::InitGoogleLogging(argv[0]);
+  gflags::ParseCommandLineFlags(&argc, &argv, true);
+
```
cramja: Also added to QuickstepCli.cpp and the NetworkIO_unittest.

cli/NetworkIO.hpp
```diff
@@ -0,0 +1,286 @@
...
+
+  DISALLOW_COPY_AND_ASSIGN(NetworkCliServiceImpl);
+};
+
+class NetworkIOHandle : public IOHandle {
```
cramja: True, both the `Handle` and `IO` implementations could be final.

cli/NetworkIO.hpp
```diff
@@ -0,0 +1,286 @@
...
+  }
+
+  IOHandle* getNextIOHandle() override {
+    return new NetworkIOHandle(service_.waitForRequest());
+  }
+
+  /**
+   * @return The underlying service which interacts with gRPC.
+   */
+  NetworkCliServiceImpl& getService() {
```
cramja: This was only being used by a test to kill the service so I changed it to:
```
  /**
   * @brief Kills the underlying gRPC service.
   */
  void killService() {
    service_.kill();
  }
```

cramja:
@zuyu Thank you for review. I addressed the minor stuff and will look into Async when I have more time this coming week.

cli/Flags.cpp
zuyu: `192.0.0.1` does not make sense. You need to review the network basis.

In `CliClientMain`, you could just [declare the flag](https://github.com/apache/incubator-quickstep/blob/master/cli/QuickstepCli.cpp#L148).

cli/NetworkIO.hpp
zuyu: If it is only used by a test, move this method to private, and add that test as a friend.

CMakeLists.txt
cramja: That makes sense. I will change it.

cli/IOInterface.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+#include <string>
+
+/**
+ * Virtual base defines a generic, file-based interface around IO.
+ */
+class IOInterface {
+ public:
+  IOInterface() {}
+
```
cramja: I removed the constructor but ran into a compiler error:
```
/Users/cramja/workspace/quickstep/incubator-quickstep/cli/NetworkIO.cpp:50:12: error: constructor for 'quickstep::NetworkIO' must explicitly initialize the base class 'quickstep::IOInterface' which does not have a default constructor
NetworkIO::NetworkIO() {
           ^
/Users/cramja/workspace/quickstep/incubator-quickstep/cli/IOInterface.hpp:62:7: note: 'quickstep::IOInterface' declared here
class IOInterface {
      ^
```
On doing the suggestion in the output, that too gave a compiler error.

zuyu:
I mean, if this class is always a base class, move the constructor in
protected session.

cramja:
After @hbdeshmukh and I's code review, we created:
* QUICKSTEP-90
* QUICKSTEP-91
in response to some of the issues we found. @zuyu 90 will address the async upgrade.

Additionally, we changed the locks in `NetworkIO.hpp` to be quickstep locks, not `std` locks.

zuyu:
We need to sync up with the master branch before merge.

PR #234, opened 2017-04-18
Add protobuf support for UNION ALL operator.

Terranlee:
This PR implements protobuf for Union All operator in distributed setting.

### DIFF
--- a/relational_operators/UnionAllOperator.cpp
+++ b/relational_operators/UnionAllOperator.cpp
@@ -65,7 +65,7 @@ void UnionAllOperator::addWorkOrdersSingleRelation(
       container->addNormalWorkOrder(
           new UnionAllWorkOrder(
               query_id_,
-              input_relations_[relation_index],
+              *input_relations_[relation_index],
               input_block_id,
               select_attribute_ids_[relation_index],
               output_destination,
@@ -75,11 +75,11 @@ void UnionAllOperator::addWorkOrdersSingleRelation(
   } else {
     std::size_t num_generated = num_workorders_generated_[relation_index];
     const std::vector<block_id> &all_blocks = input_relations_block_ids_[relation_index];
-    while (num_generated < all_blocks .size()) {
+    while (num_generated < all_blocks.size()) {
       container->addNormalWorkOrder(
           new UnionAllWorkOrder(
               query_id_,
-              input_relations_[relation_index],
+              *input_relations_[relation_index],
               all_blocks[num_generated],
               select_attribute_ids_[relation_index],
               output_destination,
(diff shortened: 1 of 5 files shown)


relational_operators/UnionAllOperator.cpp
```diff
@@ -128,14 +128,56 @@ bool UnionAllOperator::getAllWorkOrders(
 }
 
 bool UnionAllOperator::getAllWorkOrderProtos(WorkOrderProtosContainer* container) {
-  // TODO(tianrun): Add protobuf for UnionAllWorkOrder to support distributed mode.
-  LOG(FATAL) << "UnionAllOperator is not supported in distributed mode yet.";
-  return true;
+  if (!stored_generated_) {
+    for (std::size_t relation_index = 0; relation_index < input_relations_.size(); ++relation_index) {
```
cramja: I'd suggest to refactor this into a small `generateStored (?)` method, because it seems like isolatable functionality.

relational_operators/UnionAllOperator.cpp
```diff
@@ -128,14 +128,56 @@ bool UnionAllOperator::getAllWorkOrders(
...
+      const relation_id relation = input_relations_[relation_index]->getID();
+      const std::vector<attribute_id> &attributes = select_attribute_ids_[relation_index];
+      while (num_generated < all_blocks.size()) {
+        container->addWorkOrderProto(createWorkOrderProto(all_blocks[num_generated], relation, attributes), op_index_);
+        ++num_generated;
+      }
+      num_workorders_generated_[relation_index] = num_generated;
+    }
+  }
+  return stored_generated_ && done_feeding_input_relation_;
```
cramja: Is `stored_generated_ && ...` redundant, as it will always be true here?

relational_operators/UnionAllOperator.hpp
```diff
@@ -149,6 +151,11 @@ class UnionAllOperator : public RelationalOperator {
                                    InsertDestination *output_destination,
                                    const std::size_t relation_index);
 
+  // Create work order proto
```
cramja: Does this require a full doxy comment?

cramja:
LTM if comments are addressed.

relational_operators/UnionAllOperator.hpp
zuyu: No for a private method.

jianqiao:
LGTM! Merging.

zuyu:
@Terranlee Unfortunately, I observed some non-deterministic results for the following two queries in `quickstep_queryoptimizer_tests_executiongenerator_select_distributed`.

So I suggest to create a new test for `union` in the single-node mode only.

```
SELECT int_col AS result FROM test
WHERE int_col < 5
UNION
SELECT int_col + 3 AS result FROM test
WHERE int_col < 5;

SELECT float_col FROM test
WHERE int_col < 5
ORDER BY int_col DESC
LIMIT 5
UNION ALL
SELECT float_col FROM test
WHERE int_col < 5
ORDER BY int_col DESC
LIMIT 10
```

PR #239, opened 2017-04-24
Add ThreadPrivateCompactKeyHashTable for aggregation.

jianqiao:
This PR implements a new specialized aggregation hash table that is preferable for two-phase (i.e. aggregate locally and then merge) aggregation with small-cardinality group-by keys. To use this hash table, it also requires that the group-by keys have fixed-length types with total byte size no greater than 8 (so that each composite key can be packed into a 64-bit QWORD).

This PR together with #237 can improve the performance TPC-H Q01 from ~11.5s to ~5.3s, with scale factor 100 on a cloud lab machine.

### DIFF
--- a/storage/AggregationOperationState.cpp
+++ b/storage/AggregationOperationState.cpp
@@ -48,6 +48,7 @@
 #include "storage/StorageBlockInfo.hpp"
 #include "storage/StorageManager.hpp"
 #include "storage/SubBlocksReference.hpp"
+#include "storage/ThreadPrivateCompactKeyHashTable.hpp"
 #include "storage/TupleIdSequence.hpp"
 #include "storage/TupleStorageSubBlock.hpp"
 #include "storage/ValueAccessor.hpp"
@@ -69,10 +70,10 @@ namespace quickstep {
 DEFINE_int32(num_aggregation_partitions,
              41,
              "The number of partitions used for performing the aggregation");
-DEFINE_int32(partition_aggregation_num_groups_threshold,
-             500000,
-             "The threshold used for deciding whether the aggregation is done "
-             "in a partitioned way or not");
+DEFINE_uint64(partition_aggregation_num_groups_threshold,
+              100000,
+              "The threshold used for deciding whether the aggregation is done "
+              "in a partitioned way or not");
 
 AggregationOperationState::AggregationOperationState(
     const CatalogRelationSchema &input_relation,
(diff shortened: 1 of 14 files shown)


storage/ThreadPrivateCompactKeyHashTable.hpp
```diff
@@ -0,0 +1,230 @@
...
+class Type;
+
+/**
+ * @brief Specialized aggregation hash table that is preferable for two-phase
+ *        aggregation with small-cardinality group-by keys. To use this hash
+ *        table, it also requires that the group-by keys have fixed-length types
+ *        with total byte size no greater than 8 (so that the keys can be packed
+ *        into a 64-bit QWORD).
+ */
+class ThreadPrivateCompactKeyHashTable : public AggregationStateHashTableBase {
```
zuyu: Could we mark `final` so that all `override` methods would have performance benefits?

storage/ThreadPrivateCompactKeyHashTable.cpp
```diff
@@ -0,0 +1,422 @@
...
+        key_sizes_[i],
+        [&](auto key_size) -> void {  // NOLINT(build/c++11)
+      ConstructKeyCode<decltype(key_size)::value>(
+          key_code_offset, key_attr_id.attr_id, accessor, key_codes);
+    });
+    key_code_offset += key_sizes_[i];
+  }
+
+  std::vector<BucketIndex> bucket_indices;
+  bucket_indices.reserve(num_tuples);
```
jianqiao: The `reserve` approach is for avoiding the zero-initialization cost.

storage/AggregationOperationState.cpp
```diff
@@ -715,7 +719,18 @@ void AggregationOperationState::finalizeHashTable(
     finalizeHashTableImplPartitioned(partition_id, output_destination);
   } else {
     DCHECK_EQ(0u, partition_id);
-    finalizeHashTableImplThreadPrivate(output_destination);
+    DCHECK(group_by_hashtable_pool_ != nullptr);
+    switch (group_by_hashtable_pool_->getHashTableImplType()) {
+      case HashTableImplType::kSeparateChaining:
```
hbdeshmukh: It's better if we are consistent in naming the HashTableImplType and the corresponding finalize hash table function name.

hbdeshmukh:
Looks good to me.

PR #240, opened 2017-04-24
Fix the issues with the common subexpression feature.

jianqiao:
This PR fixes the issues in #237.

### DIFF
--- a/query_optimizer/rules/ExtractCommonSubexpression.hpp
+++ b/query_optimizer/rules/ExtractCommonSubexpression.hpp
@@ -111,7 +111,7 @@ class ExtractCommonSubexpression : public Rule<physical::Physical> {
   bool visitAndCount(
       const expressions::ExpressionPtr &expression,
       ScalarCounter *counter,
-      ScalarHashable *hashable);
+      ScalarHashable *hashable) const;
 
   // Traverse the expression tree and transform subexpressions (to common
   // subexpressions) if applicable.
--- a/query_optimizer/rules/ReuseAggregateExpressions.cpp
+++ b/query_optimizer/rules/ReuseAggregateExpressions.cpp
@@ -157,7 +157,7 @@ P::PhysicalPtr ReuseAggregateExpressions::applyToNode(
     count_star_ref = *it;
 
     for (++it; it != count_star_info.end(); ++it) {
-      agg_refs[*it].reset(new AggregateReference(count_star_ref));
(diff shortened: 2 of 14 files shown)


query_optimizer/rules/ReuseAggregateExpressions.cpp
```diff
@@ -194,7 +194,7 @@ P::PhysicalPtr ReuseAggregateExpressions::applyToNode(
             sum_it == ref_map.end() ? kInvalidRef : sum_it->second.front();
 
         for (const std::size_t idx : avg_it->second) {
-          agg_refs[idx].reset(new AggregateReference(sum_ref, count_ref));
+          agg_refs[idx] = std::make_unique<AggregateReference>(sum_ref, count_ref);
```
cramja: What is the difference between using the assignment and `make_unique` versus `reset` and `new`?

zuyu: Memory leaks upon system failures when using `new`.

zuyu:
@jianqiao We missed the following comments in this PR:

* [Remove unused `moveTypedValue` method in `types/containers/ColumnVector.hpp`](https://github.com/apache/incubator-quickstep/pull/237#discussion_r112827764)
* [Change `hash_cache_` as a `std::uint64_t`, instead of a smart pointer](https://github.com/apache/incubator-quickstep/pull/237#discussion_r112828399)
* [`override` destructor needed for `CommonSubexpression`](https://github.com/apache/incubator-quickstep/pull/237#discussion_r112847577)

query_optimizer/rules/ExtractCommonSubexpression.hpp
```diff
@@ -111,7 +111,7 @@ class ExtractCommonSubexpression : public Rule<physical::Physical> {
   bool visitAndCount(
       const expressions::ExpressionPtr &expression,
       ScalarCounter *counter,
-      ScalarHashable *hashable);
+      ScalarHashable *hashable) const;
```
zuyu: All methods below in this file could be marked as `const`:

* `transformExpressions`
* `transformExpression`
* `visitAndTransform`

jianqiao: These three methods involve the invoking of `optimizer_context_->nextExprId()` which modifies  the context (symbol id counter) of `optimizer_context_`.

zuyu: Good catch!

PR #241, opened 2017-04-25
Added the optimizer support for Partitioned Hash Joins.

zuyu:
This PR added the optimizer support for Partitioned Hash Joins, if one of the input relation has partition info. For repartitioning, it may add `SelectOperator` for a base relation; otherwise, we may reuse the existing `SelectOperator`, and modify the `InsertDestination` of the output relation to `PartitionAwareInsertDestination`.

```
/*
 Whether Build or Probe needs to repartition.

 --------------------------------------------------------------------------
 | Probe \ Build    | No Partition  | Hash Partition h' | Other Partition |
 --------------------------------------------------------------------------
 | No Partition     | false \ false |   true \ false    |  true \ true    |
 --------------------------------------------------------------------------
 | Hash Partition h | false \ true  | false* \ false    | false \ true    |
 --------------------------------------------------------------------------
 | Other Partition  |  true \ true  |   true \ false    |  true \ true    |
 --------------------------------------------------------------------------

 Hash Partition h / h': the paritition attributes are as the same as the join attributes.
 *: If h and h' has different number of partitions, the probe relation needs to repartition.
*/
```

Assigned to @jianqiao.

### DIFF
--- a/query_execution/QueryContext.cpp
+++ b/query_execution/QueryContext.cpp
@@ -54,6 +54,8 @@ using std::vector;
 
 namespace quickstep {
 
+constexpr QueryContext::insert_destination_id QueryContext::kInvalidInsertDestinationId;
+
 QueryContext::QueryContext(const serialization::QueryContext &proto,
                            const CatalogDatabaseLite &database,
                            StorageManager *storage_manager,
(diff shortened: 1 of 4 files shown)


zuyu:
I will revisit the current implementation so that repartition support will be added in the physical plan, instead of the execution one.

PR #244, opened 2017-05-03
Fix a problem in CopyGroupList + minor style fixes

jianqiao:
This PR adjusts the implementation of `CopyGroupList::merge_contiguous()` to handle some cases correctly.

E.g. let `R`be a split-row-store relation with integer attributes `x`,`y`,`z`. It is expected that for query
```
SELECT x, y, z
FROM R;
```
`merge_contiguous()` merges all attributes into one `ContiguousAttrs` copy group, with `bytes_to_copy` set to 12 (size of three integers).

It is also expected that for query
```
SELECT x, y, y
FROM R;
```
`merge_contiguous()` merges all attributes into two `ContiguousAttrs` copy groups, one for `x, y` and one for the last `y`.

### DIFF
--- a/storage/tests/SplitRowStoreTupleStorageSubBlock_unittest.cpp
+++ b/storage/tests/SplitRowStoreTupleStorageSubBlock_unittest.cpp
@@ -761,40 +761,41 @@ TEST_P(SplitRowStoreTupleStorageSubBlockTest, GetCopyGroupsForAttributeMapTest) 
   CopyGroupList copy_groups;
   dst_store->getCopyGroupsForAttributeMap(attr_map, &copy_groups);
 
-  std::vector<ContiguousAttrs>& contiguous_attrs = copy_groups.contiguous_attrs_;
-  std::vector<VarLenAttr>& varlen_attrs = copy_groups.varlen_attrs_;
+  const std::vector<ContiguousAttrs> &contiguous_attrs = copy_groups.contiguous_attrs;
+  const std::vector<VarLenAttr> &varlen_attrs = copy_groups.varlen_attrs;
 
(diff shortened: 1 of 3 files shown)


storage/tests/SplitRowStoreTupleStorageSubBlock_unittest.cpp
```diff
@@ -761,61 +761,89 @@ TEST_P(SplitRowStoreTupleStorageSubBlockTest, GetCopyGroupsForAttributeMapTest)
...
+              varlen_attrs[0].bytes_to_advance);
   } else {
     EXPECT_EQ(3, contiguous_attrs.size());
-    EXPECT_EQ(8, contiguous_attrs[0].bytes_to_copy_);
-    EXPECT_EQ(8 + size_of_string, contiguous_attrs[1].bytes_to_advance_);
+    EXPECT_EQ(8, contiguous_attrs[0].bytes_to_copy);
+    EXPECT_EQ(8 + size_of_string, contiguous_attrs[1].bytes_to_advance);
   }
+
+  // Extra test 1 for merging.
```
zuyu: Could we add the following comment here?

```
  // Extra test 1 for merging to simulate the following query:
  //   SELECT x, y, z FROM R;
```

Ditto for `// // Extra test 2 for merging.` below.

zuyu:
@jianqiao Still have gcc build errors.

zuyu:
Is `ane run of` in Line 49 of `SplitRowStoreTupleStorageSubBlock.hpp` a typo of `one run of`?

Also, we need to remove the last `_` in `bytes_to_advance_` in Line 72 and `bytes_to_copy_` in Line 101 of the same file.

jianqiao:
All updated.

PR #246, opened 2017-05-04
Improve disjunctive predicate pushdown to optimize more scenarios

jianqiao:
This PR improves the `PushDownLowCostDisjunctivePredicate` optimization to consider more applicable scenarios -- precisely it will push down the partial predicate when it has a low selectivity. This results in 4X speedup for TPC-H Q19.


zuyu:
LGTM!

PR #249, opened 2017-05-17
QUICKSTEP-76: Enabled LIP in the distributed version.

zuyu:
Assigned to @jianqiao.

This PR enables LIP feature in the distributed version. By computing the LIP filter group (aka equivalence class) based on the LIP filter usages in a LIP deployment, any work orders that use a LIP filter from the same group will be scheduled to the same node.

In all SSB queries, since there is only one LIP filter group, when enabling LIP feature, all work orders that use any LIP will be scheduled to one node.

Note that this PR is not compatible with partitions, until partitioned aggregations are supported.


jianqiao:
LGTM! Merging.

PR #255, opened 2017-06-05
Refactored the operator support for broadcast hash join.

zuyu:
If the build side does not have partitions, but the probe does, the build side will be broadcast to each and every partition of the probe side.


jianqiao:
LGTM! Merging.

PR #257, opened 2017-06-14
Added Partition Rule for HashJoin.

zuyu:
This PR substitutes #241 to add support for partitioned hash join. The rule adds repartitions if needed as follows:

```
/*
 * Whether left or right side needs to repartition.
 *
 * --------------------------------------------------------------------------
 * | Right \ Left     | No Partition  | Hash Partition h' | Other Partition |
 * --------------------------------------------------------------------------
 * | No Partition     | false \ false |  false \ false    |  true \ true    |
 * --------------------------------------------------------------------------
 * | Hash Partition h | false \ true  | false# \ false    | false \ true    |
 * --------------------------------------------------------------------------
 * | Other Partition  |  true \ true  |   true \ false    |  true \ true    |
 * --------------------------------------------------------------------------
 *
 * Hash Partition h / h': the partition attributes are as the same as the join attributes.
 * #: If h and h' has different number of partitions, the right side needs to repartition.
 */
```

Note that we allow `PartitionSchemeHeader` with empty partition attributes. It happens when the input partition attributes get projected out. In such case, the output partition scheme header should have the same partition id as the input, and thus avoid repartition using some random partition attributes from the project expressions.

Assigned to @jianqiao.

### DIFF
--- a/query_optimizer/PhysicalGenerator.cpp
+++ b/query_optimizer/PhysicalGenerator.cpp
@@ -31,6 +31,7 @@
 #include "query_optimizer/rules/ExtractCommonSubexpression.hpp"
 #include "query_optimizer/rules/FuseAggregateJoin.hpp"
 #include "query_optimizer/rules/InjectJoinFilters.hpp"
+#include "query_optimizer/rules/Partition.hpp"
 #include "query_optimizer/rules/PruneColumns.hpp"
 #include "query_optimizer/rules/PushDownLowCostDisjunctivePredicate.hpp"
 #include "query_optimizer/rules/ReduceGroupByAttributes.hpp"
@@ -61,6 +62,10 @@ DEFINE_bool(reorder_hash_joins, true,
             "joins. The optimization applies a greedy algorithm to favor smaller "
             "cardinality and selective tables to be joined first, which is suitable "
             "for queries on star-schema tables.");
+
+DEFINE_bool(use_partition_rule, true,
+            "If true, apply an optimization to support partitioned inputs. The "
+            "optimization may add additional Selection for repartitioning.");
 
 DEFINE_bool(use_filter_joins, true,
             "If true, apply an optimization that strength-reduces HashJoins to "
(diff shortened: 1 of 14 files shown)


query_optimizer/PhysicalGenerator.cpp
```diff
@@ -176,6 +177,8 @@ P::PhysicalPtr PhysicalGenerator::optimizePlan() {
     rules.emplace_back(new AttachLIPFilters());
   }
 
+  rules.push_back(std::make_unique<Partition>(optimizer_context_));
```
jianqiao: Is it okay to move this rule before the two LIP-related rules? Currently the LIP information may be invalidated (and lead to incorrect results) if tree transformation happens after the two LIP passes.

Meanwhile, better add some comments describing what additional information would be added into the physical nodes after this pass.

jianqiao: The other question is that -- is this rule always needed to be applied (e.g. even in single-node mode)? If not, we may have a flag or `#ifdef` to conditionally apply this rule.

jianqiao:
Good refactoring. LGTM! Except one place in `PhysicalGenerator.cpp`.

query_optimizer/PhysicalGenerator.cpp
zuyu: In the single-node version on a `NUMA` machine, we needs this partition rule. Sure, I am adding a flag to switch this rule.

Regarding the first question, if I put the partition rule before `LIP` rules, how could I process `FilterJoin` with partitioned inputs?

zuyu: I have revisited the partition rule, and think it is ok for now to put it between two LIP-related rules, as for `HashJoin` we only have the following changes to a physical plan
 1. add additional `Selections` for base tables or subquery relations, and a few non-LIP related operators like `sort` and `union all`.
 1. change output `PartitionSchemeHeader` of a Physical Plan node.

But for partitioned aggregation in a follow-up PR, we may add another aggregate operator as a finalize step. On the other hand, the original aggregation maybe substituted by a newly created aggregation with the same input node, so I put the partition rule before `AttachLIPFilters`.

zuyu:
Merging after discussing with @jianqiao.

PR #264, opened 2017-06-17
Added physical rule for partitioned aggregations.

zuyu:
The partitioned aggregation, depending on whether having the group by expression, may generate two-level aggregations (partial locally and re-aggregation finally) and an additional selection.

If having the group by expression, the output node will he partitioned.

If in the partitioned aggregation the distinct expression is not part of the group by expressions, we will not have two-level aggregations, instead, one aggregation and the partitioned input will become no partitions.

TODO: for the single-node version, we may not need two-level aggregations (aka partial-then-re-aggregation).


jianqiao:
LGTM. Merging.

PR #265, opened 2017-06-18
Added Partition Rule For NestedLoopsJoin.

zuyu:
We may use broadcast join if the left (larger) side has partitions, while the right (smaller) does not.


jianqiao:
LGTM! Merging.

PR #272, opened 2017-07-18
Ensure Physical node with partitioned inputs has partitioned output, if possible.

zuyu:
This small PR ensures that a Physical node with partitioned inputs would also produce partitioned output, if possible.

Assigned to @jianqiao.


jianqiao:
LGTM! Merging.

PR #273, opened 2017-07-20
Determine #Partitions for Aggr State Hash Table in the optimizer.

zuyu:
This PR allows the optimizer to determine the number of partitions for an aggregation state hash table, instead of getting it during the run-time from the aggregation state.

Assigned to @jianqiao.


jianqiao:
LGTM! Merging.

PR #276, opened 2017-08-03
Fixed the check failure if a query does not have profiling info.

zuyu:
Assigned to @jianqiao.

This check failure happens for a query like `create table` statements.


jianqiao:
LGTM! Merging.

PR #277, opened 2017-08-04
Fix a bug with min/max aggregation.

jianqiao:
This PR fixes the bug with min/max aggregation (that produces incorrect results) when the aggregated column is stored with **leading-zero-truncation compression**.

The bug happens because the untyped value (i.e. a `const void *` pointer) obtained from the leading-zero-truncation compressed column's corresponding `ValueAccessor` is [the address of a temporary buffer](https://github.com/apache/incubator-quickstep/blob/master/storage/CompressedColumnStoreValueAccessor.hpp#L112) -- i.e. the value pointer becomes invalid when the next `accessor->next()` gets called. [Keeping the value pointer](https://github.com/apache/incubator-quickstep/blob/master/types/operations/comparisons/LiteralComparators-inl.hpp#L546) across [multiple iterations of](https://github.com/apache/incubator-quickstep/blob/master/types/operations/comparisons/LiteralComparators-inl.hpp#L568) accessor->next() causes the problem.

The fix is to copy the value pointer's underlying value to a local variable (so that the value remains valid across iterations of accessor->next()).

### DIFF
--- a/types/operations/comparisons/LiteralComparators-inl.hpp
+++ b/types/operations/comparisons/LiteralComparators-inl.hpp
@@ -543,7 +543,11 @@ TypedValue LiteralUncheckedComparator<ComparisonFunctor,
         const TypedValue &current,
         ValueAccessor *accessor,
         const attribute_id value_accessor_id) const {
-  const void *current_literal = current.isNull() ? nullptr : current.getDataPtr();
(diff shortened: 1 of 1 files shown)


zuyu:
LGTM.

PR #280, opened 2017-08-22
Removed an unnecessary API in RelationalOperator.

zuyu:
This PR removed `informAllBlockingDependenciesMet`, as it is redundant with regards to `checkAllBlockingDependenciesMet` in the query execution engine. Note that the current implementation of `Foreman` ensure that `getAllWorkOrders` will not be called unless the dependencies are met.

Assigned to @jianqiao.


jianqiao:
LGTM! Merging.

PR #281, opened 2017-08-24
Simplified the work order generation.

zuyu:
This PR revisited the work order generation process in the query execution engine, including simplified the recursive method to `markOperatorFinished` from `processOperator`. This change improves performance for TPC-H SF-100, while simplifies the code base.

Assigned to @jianqiao, and CC'ed @hbdeshmukh.
 
tpch-sf-100-total

### DIFF
--- a/query_execution/QueryManagerBase.hpp
+++ b/query_execution/QueryManagerBase.hpp
@@ -22,6 +22,7 @@
 
 #include <cstddef>
 #include <memory>
+#include <unordered_set>
 #include <vector>
 
 #include "catalog/CatalogTypedefs.hpp"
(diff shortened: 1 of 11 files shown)


query_execution/tests/QueryManagerSingleNode_unittest.cpp
```diff
@@ -730,122 +731,4 @@ TEST_F(QueryManagerTest, TwoNodesDAGPartiallyFilledBlocksTest) {
   EXPECT_TRUE(getOperatorFinishedStatus(id2));
 }
 
-TEST_F(QueryManagerTest, MultipleNodesNoOutputTest) {
```
hbdeshmukh: This is not a far fetched scenario. Steps to reproduce: 

1. CREATE TABLE r (a INT, b INT);
2. CREATE TABLE s (c INT, d INT);
3. Select sum(a) from r, s where r.a = s.c;

zuyu: Good point!

zuyu: I've fixed this bug. Please take another look.

After that, I will squash both commits into one.

query_execution/QueryManagerBase.hpp
```diff
@@ -338,6 +285,19 @@ class QueryManagerBase {
...
   virtual bool checkRebuildOver(const dag_node_index index) const = 0;
 
+  /**
+   * @brief Check if the given operator has ever a normal work order.
+   *
+   * @param index The index of the given operator in the DAG.
+   *
+   * @return True if the operator has a normal work order, false otherwise.
+   **/
+  virtual bool hasEverNormalWorkOrders(const dag_node_index index) const = 0;
```
hbdeshmukh: Suggest to rename this function to ``hasEverProducedNormalWorkOrder`` to make more sense.

query_execution/WorkOrdersContainer.hpp
```diff
@@ -343,6 +346,18 @@ class WorkOrdersContainer {
...
   }
 
+  /**
+   * @brief Check if the given operator has ever a normal work order.
+   *
+   * @param index The index of the given operator in the DAG.
+   *
+   * @return True if the operator has a normal work order, false otherwise.
+   **/
+  bool hasEverNormalWorkOrders(const std::size_t operator_index) const {
```
hbdeshmukh: Suggest to rename this function to ``hasEverProducedNormalWorkOrder`` to make more sense.

query_execution/tests/QueryManagerSingleNode_unittest.cpp
```diff
@@ -753,99 +752,51 @@ TEST_F(QueryManagerTest, MultipleNodesNoOutputTest) {
...
-    EXPECT_TRUE(getOperatorFinishedStatus(ids[i]));
+    EXPECT_EQ(0, getNumWorkOrdersInExecution(i));
+    EXPECT_TRUE(getOperatorFinishedStatus(i));
     if (i < kNumNodes - 1) {
       EXPECT_EQ(1, operators[i + 1]->getNumCalls(MockOperator::kDoneFeedingInputBlocks));
     }
   }
 }
 
-TEST_F(QueryManagerTest, OutOfOrderWorkOrderCompletionTest) {
```
hbdeshmukh: Why is this scenario not worthy to check?

hbdeshmukh:
I have finished one pass. Will take another look once the comments are addressed.

hbdeshmukh:
Hi @zuyu 

The changes look good to me. Can you please try the following scenarios to make sure the scheduling works fine? These are some of the corner cases that I can think of: 

1. Join of two empty tables.
2. Join of two empty tables with scan on one table.
3. Join of two empty tables with scan on both tables.
4. One empty table joined with another non-empty table. 
5. One empty table joined with another non-empty table with a scan.
6. One empty table with scan joined with another non-empty table with a scan. 
7. Any variation from above with an aggregation. 
8. A non-equi join.

hbdeshmukh:
Looks good! Thanks for testing the cases.

PR #290, opened 2017-08-29
Fixed the bug that missed assigning 'num_partitions_' in CatalogRelation.

zuyu:
Assigned to @jianqiao. And thanks for reporting this bug, which would never be caught by any unit tests.


jianqiao:
LGTM. Merging.

PR #292, opened 2017-09-01
Redirect stdout and stderr in network mode.

jianqiao:
This PR redirects `stdout / stderr` to `io_handle->out() / io_handle->err()` (see [here](https://github.com/apache/incubator-quickstep/blob/fb9f856a78d947647406f661f3f3291e294bd266/cli/QuickstepCli.cpp#L309)) for transmitting the standard stream outputs to client in the network mode. It allows `quickstep_client` to use `COPY ... TO stdout ...` command to retrieve result table in CSV format.

### DIFF
--- a/cli/QuickstepCli.cpp
+++ b/cli/QuickstepCli.cpp
@@ -69,6 +69,7 @@
 #include "utility/ExecutionDAGVisualizer.hpp"
 #include "utility/Macros.hpp"
 #include "utility/PtrVector.hpp"
+#include "utility/ScopedReassignment.hpp"
 #include "utility/SqlError.hpp"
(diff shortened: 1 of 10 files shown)


cli/QuickstepCli.cpp
```diff
@@ -150,6 +152,7 @@ DEFINE_string(mode, "local",
               "same manner as with the local cli.");
 
 DECLARE_bool(profile_and_report_workorder_perf);
+DECLARE_bool(timing);
```
zuyu: No needs for this line. Already included in the header file `cli/Flags.hpp`.

zuyu:
LGTM!

PR #295, opened 2017-09-11
Printed out the partition info in QueryPlan.

zuyu:
This PR displayed the partition info in the query plan.

Assigned to @hbdeshmukh.

```
CREATE TABLE foo (attr INT) PARTITION BY HASH(attr) PARTITIONS 4;
--
TopLevelPlan
+-plan=CreateTable[relation=foo]
| +-partition_scheme_header=ProtoDescription
| | +-Property=ProtoProperty[Property=partition_type,Value=hash]
| | +-Property=ProtoProperty[Property=num_partitions,Value=4]
| | +-Property=ProtoProperty[Property=partition_attr_id,Value=0]
| +-attributes=
|   +-AttributeReference[id=0,name=attr,relation=foo,type=Int]
+-output_attributes=
  +-AttributeReference[id=0,name=attr,relation=foo,type=Int]
```

### DIFF
--- a/query_optimizer/tests/logical_generator/Create.test
+++ b/query_optimizer/tests/logical_generator/Create.test
@@ -52,6 +52,10 @@ CREATE TABLE foo (attr INT) PARTITION BY HASH(attr) PARTITIONS 4;
 --
 TopLevelPlan
 +-plan=CreateTable[relation=foo]
+| +-partition_scheme_header=ProtoDescription
+| | +-Property=ProtoProperty[Property=partition_type,Value=hash]
+| | +-Property=ProtoProperty[Property=num_partitions,Value=4]
+| | +-Property=ProtoProperty[Property=partition_attr_id,Value=0]
 | +-attributes=
 |   +-AttributeReference[id=0,name=attr,relation=foo,type=Int]
 +-output_attributes=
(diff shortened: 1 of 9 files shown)


query_optimizer/tests/logical_generator/Create.test
```diff
@@ -63,6 +67,15 @@ PARTITION BY HASH(attr1, attr2, attr3, attr4, attr5, attr6) PARTITIONS 4;
...
 +-plan=CreateTable[relation=foo]
+| +-partition_scheme_header=ProtoDescription
+| | +-Property=ProtoProperty[Property=partition_type,Value=hash]
+| | +-Property=ProtoProperty[Property=num_partitions,Value=4]
+| | +-Property=ProtoProperty[Property=partition_attr_id,Value=0]
+| | +-Property=ProtoProperty[Property=partition_attr_id,Value=1]
+| | +-Property=ProtoProperty[Property=partition_attr_id,Value=2]
+| | +-Property=ProtoProperty[Property=partition_attr_id,Value=3]
+| | +-Property=ProtoProperty[Property=partition_attr_id,Value=4]
+| | +-Property=ProtoProperty[Property=partition_attr_id,Value=5]
```
hbdeshmukh: Is it feasible to get the attribute names instead of their value? That might be easier to understand and read.

hbdeshmukh:
I have one minor comment, otherwise looks good.

query_optimizer/tests/logical_generator/Create.test
zuyu: Here we just print out the internal states.

hbdeshmukh:
Looks good.

PR #297, opened 2017-09-14
Fixed a bug in partitioned NLJ.

zuyu:
This PR fixed a bug for partitioned NLJ w/ both stored relations. Otherwise, we may miss some tuples.

Also, this PR treats the right relation as a single partition to simplify the join execution.

Assigned to @jianqiao.


jianqiao:
LGTM. Merging.

PR #298, opened 2017-09-20
Prune columns after partition rule.

zuyu:
This small PR allows to prune columns after partition rule, and thus improve the query performance by avoiding unused attributes.


jianqiao:
LGTM. Merging.

PR #299, opened 2017-09-20
QUICKSTEP-104 Fix the problem that LineReader cannot recognize a command if there are whitespaces before it.

jianqiao:
This PR fixes a bug that the Quickstep REPL cannot recognize a command (e.g. `\d`, `\analyze`) if there are whitespaces or empty lines before the command.

### DIFF
--- a/cli/LineReader.cpp
+++ b/cli/LineReader.cpp
@@ -23,6 +23,8 @@
 #include <cctype>
 #include <string>
 
+#include "glog/logging.h"
+
 using std::ispunct;
 using std::size_t;
 using std::string;
@@ -171,7 +173,7 @@ std::string LineReader::getNextCommand() {
             case '.':
             case '\\':  //  Fall Through.
               // If the dot or forward slash begins the line, begin a command search.
-              if (scan_position == 0) {
+              if (special_char_location == multiline_buffer.find_first_not_of(" \t\r\n")) {
                 line_state = kCommand;
               } else {
                 // This is a regular character, so skip over it.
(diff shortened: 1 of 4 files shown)


cli/tests/command_executor/D.test
```diff
@@ -69,6 +71,7 @@ INSERT INTO foo3 values(5, 1, 1.0, 1.0, 'XYZZ');
  col4   | Float  
  col5   | Char(5)
 ==
+
```
zuyu: In the test, is it possible to also check special chars `\t` and `\r`?

cli/LineReader.cpp
```diff
@@ -171,7 +173,7 @@ std::string LineReader::getNextCommand() {
             case '.':
             case '\\':  //  Fall Through.
               // If the dot or forward slash begins the line, begin a command search.
-              if (scan_position == 0) {
+              if (special_char_location == multiline_buffer.find_first_not_of(" \t\r\n")) {
```
zuyu: I think we should keep `scan_position == 0` as well. Otherwise, it seems that we require some special chars before a command string.

cli/tests/command_executor/D.test
jianqiao: Probably no. The test input does not go through `LineReader`.

cli/LineReader.cpp
jianqiao: Yes it should cover all the possible situations.

zuyu:
LGTM!

PR #300, opened 2017-09-21
QUICKSTEP-106: Hash-Join-Fuse: Feature added and tests modified.

dylanpbacon:


query_optimizer/rules/FuseHashSelect.cpp
```diff
@@ -0,0 +1,91 @@
...
+#include "query_optimizer/expressions/ExpressionUtil.hpp"
+#include "query_optimizer/expressions/PatternMatcher.hpp"
+#include "query_optimizer/expressions/Predicate.hpp"
+#include "query_optimizer/physical/HashJoin.hpp"
+#include "query_optimizer/physical/PatternMatcher.hpp"
+#include "query_optimizer/physical/Physical.hpp"
+#include "query_optimizer/physical/PhysicalType.hpp"
+#include "query_optimizer/physical/Selection.hpp"
+
+#include "glog/logging.h"
```
zuyu: Do we actually use sth from all these headers above? If no, please remove the unused headers.

query_optimizer/rules/FuseHashSelect.hpp
```diff
@@ -0,0 +1,67 @@
...
+ **/
+
+#ifndef QUICKSTEP_QUERY_OPTIMIZER_RULES_FUSE_HASH_SELECT_HPP_
+#define QUICKSTEP_QUERY_OPTIMIZER_RULES_FUSE_HASH_SELECT_HPP_
+
+#include <cstdint>
+#include <memory>
+#include <string>
+
+
```
zuyu: Remove this duplicated empty line, and unused headers, if any.

query_optimizer/rules/FuseHashSelect.hpp
```diff
@@ -0,0 +1,67 @@
...
+    physical::PhysicalPtr applyToNode(const physical::PhysicalPtr &node) override;
+
+ private:
+    DISALLOW_COPY_AND_ASSIGN(FuseHashSelect);
+};
+
+}  // namespace optimizer
+}  // namespace quickstep
+
+#endif /* QUICKSTEP_QUERY_OPTIMIZER_RULES_FUSE_HASH_SELECT_HPP_ */
```
zuyu: The new style is `#endif  // QUICKSTEP_QUERY_OPTIMIZER_RULES_FUSE_HASH_SELECT_HPP_`.

query_optimizer/strategy/Join.cpp
```diff
@@ -371,6 +373,7 @@ void Join::addHashJoin(const logical::ProjectPtr &logical_project,
                           left_join_attributes,
                           right_join_attributes,
                           residual_predicate,
+                          build_predicate,
```
zuyu: Optionally, I think `E::PredicatePtr()  /* build_predicate */,` is better.

zuyu:
We need to fix the performance bugs before merging.

|  **TPC-H** | **master** | **this PR** |
|  ------ | ------: | ------: |
|  01 | 5,421 | 5,039 |
|  02 | 402 | 3,968 |
|  03 | 1,229 | 3,506 |
|  04 | 2,472 | 1,242 |
|  05 | 2,807 | 3,731 |
|  06 | 399 | 394 |
|  07 | 1,754 | 3,432 |
|  08 | 906 | 9,961 |
|  09 | 6,766 | TO |

query_optimizer/strategy/Join.cpp
dylanpbacon: Agreed, less superfluous that way.

dylanpbacon:
I believe all bugs have been fixed except the LIPFilter applying to modified HashJoins. Jianqiao will be taking care of that I think. Below are the updated benchmark values.


TPC-H | Master | PR | Change
-- | -- | -- | --
1 | 5159.529375 | 5210.64075 | 0.9901909616
2 | 360.623375 | 2624.057875 | 0.1374296575
3 | 1191.425875 | 1179.699875 | 1.009939816
4 | 2468.887125 | 2231.8005 | 1.106231101
5 | 3577.305125 | 3599.183125 | 0.9939213985
6 | 382.312125 | 379.8185 | 1.006565307
7 | 1714.8465 | 1770.731 | 0.9684398703
8 | 874.823375 | 879.3375 | 0.9948664478
9 | 4803.38075 | 5012.880125 | 0.9582077828
10 | 3110.523875 | 3110.895625 | 0.9998805007
11 | 255.26625 | 241.432875 | 1.057296982
12 | 1419.048375 | 1382.390375 | 1.026517835
13 | 2947.814 | 2933.04975 | 1.005033754
14 | 742.84975 | 733.042625 | 1.013378656
15 | 426.758625 | 429.820375 | 0.9928766755
16 | 1949.653375 | 2026.087375 | 0.9622750722
17 | 1243.635125 | 992.756375 | 1.252709281
18 | 3157.806875 | 3153.086875 | 1.001496946
19 | 342.213875 | 574.703 | 0.5954621344
20 | 954.45075 | 943.952875 | 1.011121185
21 | 8464.721875 | 7656.392875 | 1.105575695
22 | 711.613125 | 727.60525 | 0.9780208774
  |   |   |  
Total | 46259.4895 | 47793.3655 | 0.9679060894

dylanpbacon:
Rebased this with master, this branch is now up to date. LIP Filters still need to be fixed to work properly with this feature.

jianqiao:
@dylanpbacon Hi Dylan, I updated the LIP related stuff and put it in this [branch](https://github.com/apache/incubator-quickstep/tree/Hash-Join-Fuse), you may just fetch the changes to reset your repo's `Hash-Join-Fuse` branch, then `git push -f`, then we can merge this PR.

**Note:**
The optimization is enabled by default, and I added an extra gflag `fuse-hash-select-threshold` that is set to one million (`1000000u`) by default. A fusion transformation is applied only when the estimated cardinality of the build-side selection is _greater_ than the threshold.

Overall speaking, the fuse-hash-select optimization is especially beneficial when _the build-side selection has large output cardinality_ (e.g. TPC-H Q21), and the benefits come from two aspects:
(1) smaller memory footprint,
(2) avoiding materialization of the selection's output.

However, due to some issues in current implementation of HashJoinOperator (and buffer manager), the fusion may slow down some queries (e.g. TPC-H Q02, 400ms -> 700ms with LIP fixed). The `fuse-hash-select-threshold` prevents those situations.

zuyu:
Hi @jianqiao Could you explain more regarding the suspect reason of the slowdown in the general case (i.e., w/o the threshold)? Thanks!

dylanpbacon:
This PR is now consistent with the changes @jianqiao made. It is ready to be merged, I believe. I do not have rights for this yet I think.

jianqiao:
@zuyu Currently `HashJoinOperator` performance is sensitive to the _number of build blocks per probe block_ due to the concurrency bottleneck within LRU policy enforcer.

Consider the situation that the build side relation has `N` blocks and the number of blocks decreases to `M` after applying the predicate, where `N` is very large but `M` is small. Then materializing the filtered build-side relation incurs only a small overhead, but it dramatically reduces the _number of build blocks per probe block_.

jianqiao:
Merging.

query_optimizer/ExecutionGenerator.cpp
```diff
@@ -966,7 +975,9 @@ void ExecutionGenerator::convertHashJoin(const P::HashJoinPtr &physical_plan) {
   const CatalogRelation *probe_relation = probe_relation_info->relation;
 
   // FIXME(quickstep-team): Add support for self-join.
-  if (build_relation == probe_relation) {
+  // We check to see if the build_predicate is null as certain queries that
+  // support hash-select fuse will result in the first check being true.
+  if (build_relation == probe_relation && physical_plan->build_predicate() == nullptr) {
```
zuyu: This change introduces a bug, because the execution engine uses the relation id to determine the build or the probe side. In the self-join case, it thus always goes to the if-case (mostly the build side), even we actually reference the probe side with the same relation id.

We do not need any further action for this, because #347 will introduce the self-join.

PR #301, opened 2017-09-27
Bug fix in LockManager loop

hbdeshmukh:
- Added a false condition for acquiring lock
- Added clarifying comment.

### DIFF
--- a/transaction/LockManager.cpp
+++ b/transaction/LockManager.cpp
@@ -80,21 +80,22 @@ void LockManager::run() {
         if (request.getRequestType() == RequestType::kReleaseLocks) {
           CHECK(releaseAllLocks(request.getTransactionId()))
               << "Unexpected condition occured.";
-
         } else if (acquireLock(request.getTransactionId(),
                                request.getResourceId(),
                                request.getAccessMode())) {
+          // Lock has been acquired.
+          LOG(INFO) << "Transaction "
+                    << std::to_string(request.getTransactionId())
+                    << " acquired " + request.getResourceId().toString();
+
+          permitted_requests_.push(request);
+        } else {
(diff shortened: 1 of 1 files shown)


transaction/LockManager.cpp
```diff
@@ -81,15 +81,17 @@ void LockManager::run() {
           CHECK(releaseAllLocks(request.getTransactionId()))
               << "Unexpected condition occured.";
 
-        } else if (acquireLock(request.getTransactionId(),
```
hakanmemisoglu: Good catch!

hbdeshmukh:
Fixed the style issues and did minor refactoring.

transaction/LockManager.cpp
```diff
@@ -79,22 +79,23 @@ void LockManager::run() {
         const LockRequest request = incoming_requests_.popOne();
         if (request.getRequestType() == RequestType::kReleaseLocks) {
           CHECK(releaseAllLocks(request.getTransactionId()))
-              << "Unexpected condition occured.";
-
+          << "Unexpected condition occured.";
```
zuyu: I think the original style is correct.

transaction/LockManager.cpp
```diff
@@ -79,22 +79,23 @@ void LockManager::run() {
...
         } else if (acquireLock(request.getTransactionId(),
-                               request.getResourceId(),
-                               request.getAccessMode())) {
+                                request.getResourceId(),
+                                request.getAccessMode())) {
```
zuyu: Please align with the first argument.

hbdeshmukh:
Fixed the style issues.

zuyu:
LGTM!

PR #303, opened 2017-09-29
Fixed the root path check in the validate_cmakelists script.

zuyu:
This small PR allows not to print out the incorrect warning, when the forked codebase root directory name does not end up with `quickstep`.


hbdeshmukh:
Looks good to me.

PR #305, opened 2017-09-29
Optimized the mod operation in HashPartition.

zuyu:
This small PR optimized out the mod operation `%` when possible.


hbdeshmukh:
Looks good to me.

PR #307, opened 2017-10-05
Moved InsertDestination::getTouchedBlocks as a private method.

zuyu:
The method is used only for testing.

Assigned to @hbdeshmukh.


hbdeshmukh:
Looks good.

PR #308, opened 2017-10-05
Removed unused argument always_mark_full.

zuyu:
This PR cleans up the unused argument `always_mark_full`, and avoids branching overhead.

### DIFF
--- a/storage/InsertDestinationInterface.hpp
+++ b/storage/InsertDestinationInterface.hpp
@@ -104,7 +104,7 @@ class InsertDestinationInterface {
    *        insertion from ValueAccessor even when partially full.
    **/
   virtual void bulkInsertTuples(ValueAccessor *accessor,
-                                bool always_mark_full = false) = 0;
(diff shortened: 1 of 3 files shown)


storage/InsertDestinationInterface.hpp
```diff
@@ -104,7 +104,7 @@ class InsertDestinationInterface {
    *        insertion from ValueAccessor even when partially full.
    **/
   virtual void bulkInsertTuples(ValueAccessor *accessor,
-                                bool always_mark_full = false) = 0;
+                                const bool always_mark_full = false) = 0;
```
hbdeshmukh: Similar to other variants of bulkInsert methods, why are we not getting rid of the ``always_mark_full`` parameter in this function?

hbdeshmukh:
Looks good.

storage/InsertDestinationInterface.hpp
zuyu: It is required by the first pass of the multi-pass sort.

PR #309, opened 2017-10-06
Added ProbabilityStore class

hbdeshmukh:
- Used to store probabilities of objects.
- Probabilities are of two kinds: Individual and cumulative.
- All the individual probabilities within the store add up to one.
- Support for finding the object with given cumulative probability.

This PR is in continuation with the set of PRs for QUICKSTEP-107.


hbdeshmukh:
Thanks for the comments. I have addressed all of them. The precision with floats is good enough for our use case.

zuyu:
LGTM. Merging.

PR #312, opened 2017-10-08
Fixed a flaky case in Catalog test.

zuyu:
This small PR fixes a flaky test case of adding an index with a different name, but the same description. However, the adding description is empty due to previous move semantics.

### DIFF
--- a/catalog/tests/Catalog_unittest.cpp
+++ b/catalog/tests/Catalog_unittest.cpp
@@ -552,13 +552,15 @@ TEST_F(CatalogTest, CatalogIndexTest) {
   IndexSubBlockDescription index_description;
   index_description.set_sub_block_type(IndexSubBlockDescription::CSB_TREE);
   index_description.add_indexed_attribute_ids(rel->getAttributeByName("attr_idx1")->getID());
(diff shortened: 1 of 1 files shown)


catalog/tests/Catalog_unittest.cpp
```diff
@@ -552,13 +552,15 @@ TEST_F(CatalogTest, CatalogIndexTest) {
   IndexSubBlockDescription index_description;
   index_description.set_sub_block_type(IndexSubBlockDescription::CSB_TREE);
   index_description.add_indexed_attribute_ids(rel->getAttributeByName("attr_idx1")->getID());
+  IndexSubBlockDescription index_description_copy;
+  index_description_copy.MergeFrom(index_description);
 
   EXPECT_TRUE(rel->addIndex("idx1", std::move(index_description)));
   EXPECT_TRUE(rel->hasIndexWithName("idx1"));
   // Adding an index with duplicate name should return false.
   EXPECT_FALSE(rel->addIndex("idx1", std::move(index_description)));
```
hbdeshmukh: By the same logic, we should create another copy for this move instruction.

hbdeshmukh:
Looks good to me.

PR #313, opened 2017-10-09
Relax the sort requirement in columnstore.

zuyu:
This small PR allows to create a relation w/ column store, while not specifying a sort attribute.

### DIFF
--- a/query_optimizer/resolver/Resolver.cpp
+++ b/query_optimizer/resolver/Resolver.cpp
@@ -687,7 +687,7 @@ StorageBlockLayoutDescription* Resolver::resolveBlockProperties(
 
   // Resolve TYPE property.
   // The type of the block will determine these:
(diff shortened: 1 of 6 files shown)


query_optimizer/resolver/Resolver.cpp
```diff
@@ -702,6 +702,7 @@ StorageBlockLayoutDescription* Resolver::resolveBlockProperties(
   } else if (type_string.compare("columnstore") == 0) {
     description->set_sub_block_type(
         quickstep::TupleStorageSubBlockDescription::BASIC_COLUMN_STORE);
+    // NOTE(zuyu): sort is optional.
```
hbdeshmukh: The variable name ``block_requires_sort`` is confusing, because the basic column store block doesn't really "require" sorting. You may rename the variable to ``block_allows_sort`` or something clearer.

PR #316, opened 2017-10-13
Support Multiple Tuple Inserts

robertclaus:
Added support for multiple tuple insert.

Previously
```
INSERT INTO <table> VALUES (<values for tuple 1>),(<values for tuple 2>);
```
would not work.  With this change it will insert both tuples into `<table>`.


jianqiao:
LGTM! Will merge after travis-ci's tests.

PR #319, opened 2017-10-24
Fixed the bug when partition w/ pruned columns.

zuyu:
This PR fixed the bug that when some column gets pruned, the partition is incorrect due to the wrong logical catalog attribute id from the output relation.

We should use the old logical catalog attribute id from *the input relation*.


jianqiao:
LGTM! Merging.

PR #320, opened 2017-10-25
Support Multiple Tuple Inserts

robertclaus:
Added support for multiple tuple insert.

Previously
```
INSERT INTO <table> VALUES (<values for tuple 1>),(<values for tuple 2>);
```
would not work.  With this change it will insert both tuples into `<table>`.


zuyu:
IMHO, this PR has a major design issue where for multi-tuple insertions, we create a bunch of `InsertOperator` and `SaveBlocksOperator`, each for tuple.

hbdeshmukh:
I agree with @zuyu. We should reconsider the design.

jianqiao:
It may just be fine before `ExecutionGenerator`, I think the fix is to revise `ExecutionGenerator::convertInsertTuple()` and do some modifications inside `InsertOperator`.

PR #321, opened 2017-11-03
Fix number of work orders generated for insert multiple tuples

robertclaus:
Inserting multiple tuples was creating multiple insert operators.  It now only creates one.

Also added unit tests for inserting multiple tuples.


jianqiao:
LGTM! Merging.

PR #326, opened 2017-11-28
QUICKSTEP-112 Get the list of referenced base relations

hbdeshmukh:
- Find the base relations that are referenced in a query.
- The referenced relations are stored in the QueryHandle.
- Separate method to just get this list of relations, without needing to fully optimize the query.

### DIFF
--- a/query_optimizer/rules/ReferencedBaseRelations.hpp
+++ b/query_optimizer/rules/ReferencedBaseRelations.hpp
@@ -1,0 +1,80 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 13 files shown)


query_optimizer/rules/ReferencedBaseRelations.hpp
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  std::string getName() const override { return "ReferencedBaseRelations"; }
+
+  TreeNodePtr apply(const TreeNodePtr &tree) override;
+
+  /**
+   * @brief Get the base relations referenced in a query.
+   */
+  const std::vector<const CatalogRelation *> getReferencedBaseRelations() const {
```
jianqiao: Better remove the beginning `const`, as the method returns a temporary object.

query_optimizer/rules/ReferencedBaseRelations.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+namespace quickstep {
+
+class CatalogRelation;
+
+namespace optimizer {
+
+class OptimizerContext;
+
+class ReferencedBaseRelations : public DFSTraversal<logical::Logical> {
```
jianqiao: Since this class overrides the `apply` method, we can just inherit `Rule<logical::Logical>`.

hbdeshmukh:
Hi @jianqiao 

Addressed your review comments. Thanks!

jianqiao:
LGTM! Merging.

PR #327, opened 2017-12-15
QUICKSTEP-113 Remove glog source code from third party

hbdeshmukh:
- glog source code is now downloaded through the download script.
- Added patches for glog.


hbdeshmukh:
In the past we had some issues with glog and xCode. Can someone volunteer to test this change using xCode?

cc: @jianqiao @pateljm

zuyu:
It works on `Mac`, but I have not tested using `XCode`.

jianqiao:
I have tested Xcode, it works! Will merge the PR later.

jianqiao:
Merging.

zuyu:
Hi @hbdeshmukh,

This PR that uses `glog 0.3.5` has a side effect that does not catch before: [it requires `cmake` version `3.0`](https://github.com/google/glog/blob/v0.3.5/CMakeLists.txt#L1), but the default `cmake` version in `Ubuntu 14.04 LTS` is `2.8.12.2`.

hbdeshmukh:
Thanks for catching this issue @zuyu. I will create a PR to bump up Quickstep's minimum cmake version.

PR #328, opened 2017-12-21
QUICKSTEP-114 Upgrade cpplint

hbdeshmukh:
- Added several missing headers.
- Modified the lint_everything file to add the command line flags.


parser/ParseString.hpp
```diff
@@ -115,4 +115,4 @@ class ParseString : public ParseTreeNode {
 
 }  // namespace quickstep
 
-#endif /* QUICKSTEP_PARSER_PARSE_STRING_HPP_ */
+#endif  /* QUICKSTEP_PARSER_PARSE_STRING_HPP_ */
```
zuyu: Suggest to `// QUICKSTEP_PARSER_PARSE_STRING_HPP_`.

storage/BasicColumnStoreValueAccessor.hpp
```diff
@@ -108,25 +109,25 @@ class BasicColumnStoreValueAccessorHelper {
   }
 
   template <bool check_null>
-  inline const void* getAttributeValue(const tuple_id tuple,
+  inline const void* getAttributeValue(const tuple_id input_tuple,
```
zuyu: To be consistent with above, how about using `tid`?

PR #331, opened 2018-02-02
Add a cmake option to handle the Travis CI timeout problem.

jianqiao:
This PR adds a cmake option `ENABLE_COMPARISON_INLINE_EXPANSION` to allow disabling of method specialization in various `Comparison`'s.

Turing the flag `OFF` will greatly reduce Quickstep compile time -- thus improving development productivity as well as fixing the Travis CI timeout problem. Note that the flag is by default `ON`, and will be [turned off](https://github.com/apache/incubator-quickstep/blob/539e1ebe09b5d1a2d86069ed1fdc6e9fb38c5ce7/.travis.yml#L80) during the Travis test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,7 +76,8 @@ before_script:
            -D CMAKE_CXX_COMPILER=$CXX
            -D CMAKE_LINKER=$CLINKER
            -D USE_TCMALLOC=0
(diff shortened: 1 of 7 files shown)


zuyu:
LGTM! Cool finding!

PR #332, opened 2018-02-07
Small adjustments in star schema cost model for # distinct values estimation

jianqiao:
This PR has a small adjustment in star schema cost model for # of distinct values estimation, together with a fix to a potential bug with `impliesUniqueAttributes`.

The adjustment is likely to improve query plans _when table stats are not present_. It does not affect SSB/TPC-H performance.


query_optimizer/cost_model/StarSchemaSimpleCostModel.cpp
```diff
@@ -493,7 +493,7 @@ std::size_t StarSchemaSimpleCostModel::getNumDistinctValues(
       return stat.getNumDistinctValues(rel_attr_id);
     }
   }
-  return estimateCardinalityForTableReference(table_reference);
+  return estimateCardinalityForTableReference(table_reference) * 0.1;
```
zuyu: Could you please explain more regarding why `0.1` would be a good guess for the number of distinct values?

jianqiao: This estimation ratio can be any decimal number that is not close to `1` -- in that case the optimizer would choose bad plans in some situations as the column appears to have "unique" values.

`0.1` tends to be a reasonable choice -- we may also have `0.05`, `0.2`, etc., which can be adjusted later when there are actual demands.

zuyu:
LGTM!

PR #333, opened 2018-02-19
Fix SeparateChainingHashTable::resize()

jianqiao:
This PR fixes the problem that Quickstep hangs when resizing `SeparateChainingHashTable` during the execution of `BuildHashOperator`.

Here is a sequence of queries that reproduce the problem:
```
CREATE TABLE r(x INT, y INT);
CREATE TABLE s(x INT, y INT);
CREATE TABLE t(x INT, y INT);

INSERT INTO r SELECT 1, 1 FROM generate_series(1, 200) AS g(x);
INSERT INTO s SELECT 1, 1 FROM generate_series(1, 200) AS g(x);
INSERT INTO t SELECT 1, 1 FROM generate_series(1, 1000) AS g(x);

\analyze

SELECT COUNT(*) FROM r, s, t WHERE r.x = s.x AND r.y = s.y AND s.x = t.x AND s.y = t.y;
```

The problem is caused by the [`resize()` call](https://github.com/apache/incubator-quickstep/blob/master/storage/HashTable.hpp#L1514) in `HashTable::putValueAccessorCompositeKey()` when `using_prealloc` is true. In this case, pre-allocation decides to resize the hash table in order to consume all the tuples from the current value accessor. However, `resize()` will always abort if the hash table is not "actually full", causing infinite loops.

Note that `SimpleScalarSeparateChainingHashTable` does not have the same problem, as its [`isFull` method](https://github.com/apache/incubator-quickstep/blob/master/storage/SimpleScalarSeparateChainingHashTable.hpp#L241) already takes `extra_buckets` into consideration.

Also note that `LinearOpenAddressingHashTable` seems to have avoided the hanging problem by using a [`retry_num` check](https://github.com/apache/incubator-quickstep/blob/master/storage/LinearOpenAddressingHashTable.hpp#L1203).

### DIFF
--- a/storage/SeparateChainingHashTable.hpp
+++ b/storage/SeparateChainingHashTable.hpp
@@ -243,9 +243,10 @@ class SeparateChainingHashTable : public HashTable<ValueT,
                                         HashTablePreallocationState *prealloc_state);
 
   // Determine whether it is actually necessary to resize this hash table.
-  // Checks that there is at least one unallocated bucket, and that there is
+  // Checks that there are sufficient unallocated buckets, and that there are
   // at least 'extra_variable_storage' bytes of variable-length storage free.
-  bool isFull(const std::size_t extra_variable_storage) const;
+  bool isFull(const std::size_t extra_buckets,
+              const std::size_t extra_variable_storage) const;
 
   // Helper object to manage key storage.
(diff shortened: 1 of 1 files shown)


zuyu:
LGTM!

PR #334, opened 2018-02-26
Fix iwyu include path

jianqiao:
This PR fixes the third-party library include paths for the iwyu (include-what-you-use) tool.


hbdeshmukh:
Hi @jianqiao 

This looks good to me. Have you been actively using the IWYU library? I stopped using it a while ago because of the false positives. I am not sure if it has improved.

zuyu:
Merged.

PR #335, opened 2018-02-28
DO NOT MERGE: Dockerfile Support for Distributed Deployment

yuanchenl:


zuyu:
@yuanchenl Please review and minimize your changes (i.e, using a stable version of `cmake` instead of `rc` version, why installing `grpc` twice (once using `apt-get` and built from source code), and so does `protobuf`).

Does this PR achieve the goal that by starting a container specified by the modified Dockerfile, `quickstep_cli_shell` is running as a server to accept queries from some input interface (say browser)?

yuanchenl:
The repeated installations are removed. CMake version is changed to 3.10.2 (stable).
The goal of this PR is to simply get the distributed quickstep correctly built in a docker image. 
Since different container instances have different roles (conductor, cli, executor) and different configurations, I am currently running the image using pseudoTTY (docker run -ti) and manually specify the role and the conductor address for quickstep_distributed_cli_shell inside the tty.
Do you want me to change the dockerfile so that we can specify those information during building phase and run the image without the stdin attached?

zuyu:
Before merging this PR, please rebase this branch with the master.

yuanchenl:
Rebased

PR #336, opened 2018-03-01
Fixed the bug that Executor / Cli does not create directory for StorageManager.

zuyu:
Assigned to @jianqiao.

@yuanchenl Please test this PR using Docker. Thanks!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -792,6 +792,7 @@ if (ENABLE_DISTRIBUTED)
                         glog
                         quickstep_catalog_CatalogRelation
                         quickstep_cli_Constants
+                        quickstep_cli_DefaultsConfigurator
                         quickstep_cli_Flags
                         quickstep_cli_LineReader
                         quickstep_cli_PrintToScreen
--- a/cli/DefaultsConfigurator.cpp
+++ b/cli/DefaultsConfigurator.cpp
@@ -41,10 +41,7 @@ using std::string;
 
 namespace quickstep {
 
-void DefaultsConfigurator::InitializeDefaultDatabase(const string &storage_path, const string &catalog_path) {
-  // TODO(jmp): Refactor the code in this file!
-  LOG(INFO) << "Initializing the database, creating a new catalog file and storage directory";
-
+void DefaultsConfigurator::CreateDirectory(const std::string &storage_path) {
   // Create the directory
   // TODO(jmp): At some point, likely in C++-17, we will just have the
   //            filesystem path, and we can clean this up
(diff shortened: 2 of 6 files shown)


yuanchenl:
Sorry for the late reply. 
Seg fault encountered when I am trying to copy content from tbl to an existing table.
Here's the output of the executor:
cxxdev@shiftboss-0:~/quickstep/build$ GLOG_logtostderr=1 ./quickstep_distributed_cli_shell -role=Executor -use_ethernet_ip -tmb_server_ip=128.104.222.176
I0306 05:59:10.842419   265 CalculateInstalledMemory.cpp:42] Total installed memory is 168777637888 bytes
I0306 05:59:10.844399   265 DataExchangerAsync.cpp:143] DataExchangerAsync Service listening on 128.104.222.169:46848
I0306 05:59:10.846756   265 InputParserUtil.cpp:51] Empty worker affinities provided, switching to default worker affinities
--num_workers is 40, but only specified 0 CPU affinities with --worker_affinities. 40 workers will be unaffinitized.
I0306 05:59:10.902333   265 StorageManager.cpp:238] StorageManager with Client 5 starts with Domain 2
I0306 05:59:10.902842   265 Shiftboss.cpp:100] Shiftboss TMB client ID: 6
I0306 05:59:10.960115   265 QuickstepDistributedCli.cpp:90] executor is ready
I0306 06:01:52.119858   270 WorkOrderFactory.cpp:555] Creating TextScanWorkOrder for Query 8 in Shiftboss 0
Segmentation fault

To replicate the error, please load the region tbl from the tcp-h using "copy region from '[path/to/region.tbl]' with (delimiter '|');"
The region.tbl file I used is generated by the "dbgen -s 10" command.

yuanchenl:
The seg fault of the executor happens when copy command is used.
But I didn't encounter the seg fault when I am using regular insertion

zuyu:
@yuanchenl #337 should avoid the segfault. But the root cause of the segfault is that we do not have a distributed FS for the tbl files. A workaround is that copying all tbl files to every node except for `cli`.

On the other hand, to test any new features, we need to build `Quickstep` again inside `Docker` container. Is this what you do?

yuanchenl:
Yep. It was built again in docket container.
About the testing, I wrote a script that parses the tbl into insertion command. And currently I’m testing using this workaround. 
So far so good under the tcph benchmark. I’ll generate a larger sample and test it again soon.

zuyu:
Distributing the tbl files to every data node is a better approach for large data files. On the other hand, `InsertTuples` has a lot overhead by attribute type checks and tuple copy costs.

yuanchenl:
Right... 
Though it takes awhile to copy the data between nodes, it's probably still faster for the low overhead. 
I'll test #337 ASAP.

PR #337, opened 2018-03-06
Check File Handle in TextScanWorkOrder to avoid segfault.

zuyu:
This PR avoids the segfault mentioned [here](https://github.com/apache/incubator-quickstep/pull/336#issuecomment-370674172) when opening an input file that does not exist.


yuanchenl:
It works perfect

PR #338, opened 2018-03-13
Fixed the gRPC Problem for Data Exchange

yuanchenl:
Fixed the data exchange problem encountered in distributed quickstep.

It turned out that problem of data exchange is caused by the gRPC channel's limit on the maximum receive size (4 MiB, 4194304 Bytes).

In DataExchange.proto
`message PullResponse {
  bool is_valid = 1;
  uint64 num_slots = 2;
  bytes block = 3;
}`
the block size is 4MiB (which is fine), but there are 9 bytes of overhead (num_slots 8 bytes and is_valid 1 byte). These extra bytes are the essential reason that the data exchange failed.
In this commit, I set the limit of maximum receive size to 4194213 (4MiB + 9bytes), which is right enough for us. 
With this commit, the distributed quickstep is able to pass ssb test with a scale factor of 10

### DIFF
--- a/storage/StorageManager.cpp
+++ b/storage/StorageManager.cpp
@@ -687,8 +687,13 @@ StorageManager::BlockHandle StorageManager::loadBlockOrBlob(
 #ifdef QUICKSTEP_DISTRIBUTED
     const string domain_network_address = getPeerDomainNetworkAddress(BlockIdUtil::Domain(block));
     DLOG(INFO) << "Pulling Block " << BlockIdUtil::ToString(block) << " from " << domain_network_address;
+
+    // Customize the grpc channel
+    grpc::ChannelArguments channelArgs;
+    channelArgs.SetMaxReceiveMessageSize(kGrpcChanelSize);
+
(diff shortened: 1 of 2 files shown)


zuyu:
LGTM!

storage/StorageManager.cpp
```diff
@@ -687,8 +687,13 @@ StorageManager::BlockHandle StorageManager::loadBlockOrBlob(
 #ifdef QUICKSTEP_DISTRIBUTED
     const string domain_network_address = getPeerDomainNetworkAddress(BlockIdUtil::Domain(block));
     DLOG(INFO) << "Pulling Block " << BlockIdUtil::ToString(block) << " from " << domain_network_address;
+
+    // Customize the grpc channel
+    grpc::ChannelArguments channelArgs;
+    channelArgs.SetMaxReceiveMessageSize(kGrpcChanelSize);
```
zuyu: Please add a condition wrapper based on the `GRPC` version number for this API call, since not every version has this API.

PR #339, opened 2018-04-02
Upgrade cmake version.

hbdeshmukh:
Upgrading CMake, as is required by glog which is a third party dependency (discussed in #327).


jianqiao:
LGTM! Merging.

PR #341, opened 2018-04-20
Used MergeFrom instead of CopyFrom.

zuyu:


hbdeshmukh:
What's the difference between the two functions?

PR #342, opened 2018-04-22
Quickstep-119: Added the rule that eliminates a node w/ an empty table.

zuyu:
This PR adds an optimization rule that eliminates a physical node with an empty (determined by stats) `TableReference` into a `Selection` on a temp table.


jianqiao:
LGTM! Merging.

PR #343, opened 2018-04-27
Fix all CMakeLists.txt for automated processing

jianqiao:
This PR fixes and adjusts the style of all `CMakeLists.txt` so that they become stable (i.e. well-formatted) to be processed by an automated tool.

The above mentioned tool will be proposed in a subsequent PR. It is intended to help improve developer productivity as it scans source code dependencies and automatically fixes cmakelists.


zuyu:
LGTM!

PR #346, opened 2018-04-28
QUICKSTEP-124: Add a python script to auto fix CMakeLists files

jianqiao:
This PR adds a script that intends to help improve developer productivity by automatically fixing `CMakeLists.txt` files for the Quickstep project (with best effort).

The script will do the following things:
- Scan the repo's subdirectories and collect `#include` information from all source code files.
- Parse existing `CMakeLists.txt` files and convert all "recognized" commands into proper intermediate representations -- the "unrecognized" part will be kept as "verbal" lines.
- Resolve subdirectories, targets and link dependencies. Add / delete / update the corresponding entries.
- Convert the intermediate representations back to `CMakeLists.txt` files.

**NOTE:** Currently the script is at its initial stage and will not update tests or conditional targets (i.e. those within cmake `if` commands). It is likely to work well if you just create/delete some files or add/remove some `#include`'s -- otherwise additional manual fixes may need to be done after applying the script.

### DIFF
--- a/autofix_cmakelists.py
+++ b/autofix_cmakelists.py
@@ -1,0 +1,2054 @@
+#!/usr/bin/env python
+
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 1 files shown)


zuyu:
LGTM!

PR #347, opened 2018-04-30
QUICKSTEP-121: Added the self-join support.

zuyu:
This PR added the self-join support by assigning the join side info, instead of relying on the relation id.

Note that now the join result pair of all `HashJoin`s (except for `AntiJoin`) is in the format of `<probe-tid, build-tid>`, consistent with `NestedLoopsJoin` (<left-tid, right-tid>).

Some benchmark results (10 warm runs):

```
CREATE TABLE s (x int, y int);
INSERT INTO s SELECT i, i FROM gererate_series(1, 100000000) AS gs(i);
SELECT COUNT(*) FROM s s0, s s1 WHERE s0.y = s1.x AND s0.y != -1;
```
chart-3

On the other hand, the following self-join query takes about 200 ms if using `LIP`:
```
SELECT COUNT(*) FROM s s0, s s1 WHERE s0.y = s1.x;
```

### DIFF
--- a/expressions/ExpressionFactories.cpp
+++ b/expressions/ExpressionFactories.cpp
@@ -52,6 +52,8 @@
 
 namespace quickstep {
 
+namespace S = serialization;
+
 class Type;
 
 Predicate* PredicateFactory::ReconstructFromProto(const serialization::Predicate &proto,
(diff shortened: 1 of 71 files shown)


zuyu:
Holds util PR #348 merges.

jianqiao:
LGTM! Merging.

Note that the `concretize` signature in `Physical` plans looks somehow cumbersome, we may add a `SubstitutionContext` class to wrap these in a future PR.
```
::quickstep::Predicate* concretize(
    const std::unordered_map<ExprId, const CatalogAttribute*> &substitution_map,
    const std::unordered_set<ExprId> &left_expr_ids = std::unordered_set<ExprId>(),
    const std::unordered_set<ExprId> &right_expr_ids = std::unordered_set<ExprId>()) const override;
```

PR #349, opened 2018-05-03
Fixed the bug regarding EliminateEmptyNode on an aggregate query.

zuyu:
This PR fixed a bug introduced in #342 that incorrectly reduces an aggregate query on an empty table into a temp table scan. Ironically, `quickstep_cli_tests_commandexecutor_analyze` does not proof the correctness of the rule because the `\analyze` command does not set stats due to the bug mentioned before.

In other words, we could easily reproduce the bug:
```
CREATE TABLE r (a int);
\analyze r
SELECT COUNT(*) FROM r;  // this produce nothing (in 0 row), instead of a ZERO (in 1 row).
```

This fix transforms an aggregate query correctly, and adds more tests.

### DIFF
--- a/cli/tests/command_executor/Analyze.test
+++ b/cli/tests/command_executor/Analyze.test
@@ -27,11 +27,38 @@ INSERT INTO t VALUES(0, 0);
 --
 ==
 
+SELECT COUNT(*) FROM r;
+--
++--------------------+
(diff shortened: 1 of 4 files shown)


jianqiao:
LGTM! Merging.

PR #350, opened 2018-05-04
Fixed the bug regarding EliminateEmptyNode and InsertSelection.

zuyu:
This PR fixed a bug regarding `EliminateEmptyNode` and `InsertSelection` that could be easily reproduced as the following:
```
CREATE TABLE temp (a int);
\analyze
INSERT INTO temp SELECT i FROM generate_series(1, 100) AS gs(i);
```

The problem is that `InsertSelection` allows the destination table to be empty, but the rule does not catch that.


jianqiao:
LGTM! Merging.

PR #351, opened 2018-05-05
Use Exactness info in Catalog stats.

zuyu:
When we get stats info, we should not ignore the `exactness` flag.

This PR checks the flag before answer whether a stat exists.


jianqiao:
The stats can be used to provide an estimation even when they are not exact.

PR #352, opened 2018-05-05
QUICKSTEP-125: Fixed the non-determinism in JoinReordering.

zuyu:
This PR fixed the bug in join reordering due to the pointer address-based comparison. It also added a new rule that prefers the build side which does not produce output, and did minor refactors for the file.

Assigned to @jianqiao.


jianqiao:
LGTM! Merging.

PR #354, opened 2018-05-08
Fixed the union-all elimiation case where some project expression is not an AttributeReference.

zuyu:
This PR fixed for the case where the only one child of the union-all query is non-empty, and the child has a non-AttributeReference project expression:

```
SELECT 1 AS a FROM not_empty
UNION ALL
SELECT 0 AS a FROM empty;
```


jianqiao:
LGTM! Merging.

PR #357, opened 2018-05-21
Fixed the command execution bug in the distributed version.

zuyu:
This PR fixed the command execution bugs in the distributed version, and now it is consistent with the single-node version.


jianqiao:
LGTM!

PR #358, opened 2018-06-03
Fix a bug in HashJoinOperator

jianqiao:
This PR fixes a bug in `HashJoinOperator` w.r.t. the swapping of probe/build sides in a previous PR.

### DIFF
--- a/autofix_cmakelists.py
+++ b/autofix_cmakelists.py
@@ -19,6 +19,7 @@
 
 import os
 import re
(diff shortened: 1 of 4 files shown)


zuyu:
Hold for adding tests to cover these buggy cases to prevent future bugs.

jianqiao:
Tests added. The problematic code would fail on the added tests.

